Issue #27, opened 2014-11-18
Bulk/array inserts


techfort:
bulk inserts should work.

PR #28, opened 2014-11-19
add bulkInsert

GKuChan:
#27


techfort:
This PR is now obsolete as bulk inserts are supported by the insert method. Thanks for submitting in any case!

PR #57, opened 2015-01-08
Resolves  #56

### LINKED ISSUE #56: "Data is undefined" thrown when using loadDatabase in browser
To verify that this was an issue, I made a simple index.html file w/ the following code (taken from the docs)

```javascript
    var db = new loki('test', { env: 'BROWSER'}),
            db2 = new loki('test', { env: 'BROWSER'});

    var users = db.addCollection('users');
    users.insert({
        name: 'joe'
    });
    users.insert({
        name: 'john'
    });
    users.insert({
        name: 'jack'
    });
    console.log(users.data);
    db.saveDatabase();

    db2.loadDatabase(function () {
        var users2 = db2.getCollection('users')
        console.log(users2.data);
    });
```

`loadDatabase` breaks in the following code snippet (which exists in two areas):

```javascript
        if (localStorageAvailable()) {
          self.loadJSON(localStorage.getItem(this.filename));
          cFun(null, data);
        } else {
          cFun(new Error('localStorage is not available'));
        }
```

Data is undefined and an error is thrown. I've verified that removing data from the callback resolves the issue.

chimon2000:
The following fix resolves the issue with loadDatabase not working on the browser.


obeliskos:
Good catch, thanks!

PR #67, opened 2015-01-28
Add suport to compare  null and undefined.

bnadim:
Solve a bug : 
When trying fo find with an indexed query, if value was undefined or null for some objects, calculate range would return wrong range.


obeliskos:
This change involves the most performance-sensitive code path, so i would like a chance to do some testing with it to understand the problem, performance implications and travis ci unit test failure before committing.

Thanks for spotting this i should be able to review (possibly tomorrow).

bnadim:
Hi thanks for your quick reply. Travis now passes. I made a little update to use it in comparaison operators also.
Here is the bug that i encoutered : 
coll.insert({ "key1": "value1"})
coll.insert({ "key1": "value1", "key2": "1" })
coll.find({"key2": "1"}) with an index on key2 -> this did not return any result.

bnadim:
And this is coherent with mongo way of working : http://docs.mongodb.org/manual/reference/method/cursor.sort/

bnadim:
In my latest commit, i solved another bug with operator $lt and indexed query. If there was only one value in the collection, $lt wouldn't not return anything on indexed fields.

obeliskos:
Well done, patch looks good, covers important edge cases, and smallish performance hit seems worth it.  Thanks!

Issue #76, opened 2015-02-11
"indices" as a string is splited like an array


Badisi:
[Documentation](http://lokijs.org/#/docs#addCollection) on addCollection clearly says:
```indices: an array of strings indicating indexed fields.```

but in the [wiki](https://github.com/techfort/LokiJS/wiki) the following syntax is also suggested (ie. a simple string):
```javascript
var users = db.addCollection('users', { indices: 'email' });
```

Using the second option will persist the following:
```json
"binaryIndices": {
      "e": {
        "name": "e",
        "dirty": true,
        "values": []
      },
      "m": {
        "name": "m",
        "dirty": true,
        "values": []
      },
      "a": {
        "name": "a",
        "dirty": true,
        "values": []
      },
      "i": {
        "name": "i",
        "dirty": true,
        "values": []
      },
      "l": {
        "name": "l",
        "dirty": true,
        "values": []
      }
    }
```

techfort:
This is an error in the wiki, which I will fix immediately. `indices` in Loki > 1.1 is a property of options i.e. `addCollection('users', { indices: ['email', 'phone']})`. Thanks for catching this!

Badisi:
You are welcome!

Speaking of that, if I use the following, I got an **undefined is not a function** error:
```javascript
addCollection('users', { indices: ['id'] })
```

I did not investigate it further but I think it's related to #30.
Do you an idea of what it might be ?

techfort:
To avoid clashing properties we renamed the main index in LokiJS to $loki. So unless your properties have an `id` field there may be some error thrown. Could you create a gist or a plunker so I can examine?

Badisi:
Yes, I was not trying to use the id provided by LokiJS.
The 'id' in this case is a property of a User object.
I'll see if I can provide a reproducible example.
Thanks

Badisi:
I narrow down the issue and find out that it was due to my misunderstanding of 'Collection.findOne'.
I was expecting a document to be returned instead of a ResultSet.
As we are asking for only one element, a ResultSet is kind of useless don't you think ?

Also I was expecting 'Collection.data()' to indeed return the data but instead I got an empty array.
Maybe find a bug here.

Plunker: http://plnkr.co/edit/an4zfquYf6kl1hWPRMcX?p=preview

techfort:
Ok I will close this as #78 is being worked on.

PR #77, opened 2015-02-11
Fix for #76

Badisi:
#76 

(I'll let you decide whether you wan't to allow string / array or both as an option and update this pull request and documentation accordingly)


techfort:
Thanks for this, I'd better update the docs as well, seeing as indices can now take a single string or an array of strings.

PR #101, opened 2015-03-15
add amd and requirejs support to Loki Angular Adapter

hthetiot:
I had an issue using Loki Angular Adapter in my requirejs stack cause lokijs do not export root.loki when loaded using requirejs and that was require for the adapter.

This push request is fixiing that on top add comments and trimming whitespace. 
Also move global imports collision workaround into lokijs.js instead of angular adapter.

PR #102, opened 2015-03-15
fix global imports collision workaround #101

hthetiot:


techfort:
Hey Harold,

i'm going to have to rollback a few changes you made. I like the idea of loki being transferred to a window.thirdparty object, however that's going to break the apps of all those who are instantiating a loki db by just calling `new loki` as they would now have to call `new thidParty.loki`. I'd rather add more detectors so loki gets moved to thirdParty in case angular or some other framework is used. Does that sound ok to you?

techfort:
Btw, i left loki as a property of thirdparty, then that way if the user wants to manually delete window.loki and keep it only in thirdParty they can.

hthetiot:
Leaving window.loki works for me.
I was actually wondering why it was deleted into the angular adapter originally.
Thanks for the update.

hthetiot:
If you want me to re push with keeping window.loki just ask.

techfort:
No problem and thanks for helping out on this!

PR #117, opened 2015-03-24
$containsAny

KimSchneider:
The following case: Our objects have a tags array and I want every object that has at least one of find({ tags : { '$containsAny': [ 'tag1', 'tag2' ]}

The same applys to views.

If this doesn't meet your requirements I am open for a workaround :)

### DIFF
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -141,6 +141,38 @@
       $in: function (a, b) {
         return b.indexOf(a) > -1;
       },
+      
+      $containsAny: function (a, b) {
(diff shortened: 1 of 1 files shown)


KimSchneider:
I am not really sure if I already created this pull request und you declined it, because I couldn't find any notification about that.

techfort:
Merged it a few minutes ago, did you get the notification? All good anyway!

KimSchneider:
Received this notification but not the last one, thanks anyways!

Issue #107, opened 2015-03-19
Enhancement: Browser compatibility badge/tests.


tecfu:
Hey Joe - love your project and really hope you succeed big.

I've just come across Loki in the last hour or so. A couple ideas:

- Browser compatibility badges would help new users get a sense of how well supported Loki is in the browser. https://ci.testling.com/ might be a good option. Its a bit hypocritical for me to mention this, because I haven't done it for my own projects. But I think it would benefit you.

- Add an "onChange" event for the Changes API

- I bet all the cool hipster kids in Nothern California would love you if you created a Loki/Flux Store object.

Just some thoughts. Hope this blows up.

techfort:
Hey there, thanks a lot for the kind words! I was not aware of testling.com  i will definitely look into getting browser compatibility badges. 
Adding that "onChange" should be a straight forward and brilliant idea, will get on it asap.
I also admit my complete ignorance of Flux. I know what it is but never worked with it. I will take this opportunity to expand my knowledge on the subject!
Thanks a lot again for the very constructive feedback!

jrhicks:
@tecfu @techfort - I've been thinking about how replication and lokijs could fit into the flux pattern quite a bit lately.  After my first implimentation I can say that flux slows down replicating 30,000 records .... I'm looking into some "almost flux" patterns and after facebook release their relay diagram today https://facebook.github.io/react/blog/2015/03/19/building-the-facebook-news-feed-with-relay.html ... I'm guessing that they don't mind breaking pure-flux.

Its a bit of a work in progress, but here is the diagram I have so far - http://screencast.com/t/r0B2XYtSZx

techfort:
@tecfu  @jrhicks I am going to spend the next few days familiarizing with flux. I suppose i dodged it all along because i was otherwise occupied but i definitely need to face the beast now. Nonetheless, if you feel like keeping this going with further information I will very much appreciate it.

PR #122, opened 2015-03-26
RFC: Make the tests run in a browser

roryf:
The changes here are a total hack to just get something working, I'm looking for feedback on whether this is worthwhile and what the best approach might be if so.

Following on from #107, I'm also looking at loki from a browser perspective and want to ensure it's browser support and performance is adequate. I've hacked the existing unit tests to run with Jasmine and Karma, excluding those that don't make sense in a browser context. Barring the `Collection.mode` test these all pass in:

* Chrome 41.0.22772 (OS X)
* Firefox 34 (OS X)
* Safari 8.0.3 (OS X)
* IE 11 (Win8.1)
* IE 9 (Win7)
* Firefox 35 (Win8.1)

I also hacked the `benchmark.js` to run in a browser (via `benchmark.html`) and got the following results in Chrome 41 on a MacBook Air:

```
load (insert) : 44.21ms (113096.58) ops/s
coll.get() : 816.39ms (2449809.53) ops/s

-- Benchmarking query on NON-INDEXED column --
coll.find() : 672.89ms (29722.54 ops/s) 20000 iterations
resultset chained find() :  632.06ms (31642.57 ops/s) 20000 iterations
loki dynamic view first find : 644.69ms (31022.66 ops/s) 20000 iterations
loki dynamic view subsequent finds : 4.98ms (4016064.26 ops/s) 20000 iterations

-- ADDING BINARY INDEX to query column and repeating benchmarks --
coll.find() : 532.87ms (750652.13 ops/s) 400000 iterations
resultset chained find() :  419.66ms (714864.41 ops/s) 300000 iterations
loki dynamic view first find : 515.01ms (582512.96 ops/s) 300000 iterations
loki dynamic view subsequent finds : 64.02ms (4686035.61 ops/s) 300000 iterations
```

So, assuming this is worthwhile, any comments at this stage? I was looking at how PouchDB manage their tests and there is a wide range of different test suites for different environments. So clearly running the same tests on browser and node isn't the goal, but getting something working should be (IMO).

### DIFF
--- a/benchmark.html
+++ b/benchmark.html
@@ -1,0 +1,2 @@
+<script src="src/lokijs.js"></script>
+<script src="benchmark/benchmark.js"></script>
--- a/benchmark/benchmark.js
+++ b/benchmark/benchmark.js
@@ -1,11 +1,11 @@
-var loki = require('../src/lokijs.js'),
-	db = new loki('perftest'),
+// var loki = require('../src/lokijs.js'),
+var db = new loki('perftest'),
     samplecoll = null,
     arraySize = 5000,			// how large of a dataset to generate
     totalIterations = 20000,	// how many times we search it
     results = [],
 	getIterations = 2000000;	// get is crazy fast due to binary search so this needs separate scale
-  
+
 function genRandomVal()
 {
     var text = "";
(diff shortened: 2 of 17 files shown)


techfort:
Hey Rory,

this looks great however the travis build failed, see [this](https://s3.amazonaws.com/archive.travis-ci.org/jobs/55938398/log.txt). Could you take a look and let me know? I have been considering moving over to karma so this is great timing.

roryf:
Fixed now. Like I said, this is a total hack just to prove that it would actually work.

techfort:
Great, thanks a lot! I'm hoping to get browser badges on the back of this process!

roryf:
@techfort erm, okay. Did I mention this was a total hack?

techfort:
yeah you did, but you also kickstarted the browser-compatibility testing process which i was feeling terribly lazy about, now I can just pick up from here. I wrote [gordian](https://github.com/techfort/gordian) as a spartan way to test Loki - but now that Loki has grown so much I'm going to have to ditch gordian in favour of more accepted practices (karma, mocha,jasmine, whatever..). I'll start hacking those tests myself (e.g. replace suite.assertEqual with expect()-type statements) and see if stuff still works. That's what I really meant with the above comment.

PR #129, opened 2015-04-04
Use package-quality shield (mini-badge).

alexfernandez:
Keep with the visual style of other mini badges.


techfort:
Thanks Alex!

PR #132, opened 2015-04-07
Replaced Gulp by Npm scripts

seriousme:
As discussed: replaced Gulp by npm scripts, e.g.:
```
npm run build
```
will now build the minified versions in /build
I did not link build to testing yet.
this can be done by adding:
```
"pretest":"npm run build",
```
or
```
"postbuild":"npm test",
```
to package.json and changing the karma.conf to point to  'dist/lokijs.min.js'

I also did not add browserify yet as the current version of Lokijs itself is already browserified.

Do we still need the Makefile ?


techfort:
Cool - thanks for this. We don't need makefile, i thought i removed it. Also would be nice to include a note on npm build/test in the readme or the wiki.

seriousme:
Ok, will remove the makefile

I'll update the wiki later on.


Techfort schreef op 7-4-2015 om 20:04:

PR #133, opened 2015-04-07
Reworked test framework to enable nodejs testing again

seriousme:
Coverage testing for nodejs not back in yet
Makefile removed


techfort:
Awesome stuff!

PR #135, opened 2015-04-09
Changes API removing bug

Shay12tg:
I get "Uncaught SyntaxError: Unexpected token u " after removing while using the changes API.
That is because JSON.parse is trying to parse an undefined object.


techfort:
Thanks for this!

PR #137, opened 2015-04-10
add publish to package.json

seriousme:
As requested:

Package.json now holds a few extra script lines, most importantly publish
Invoke it using 'npm run publish', if all goes well it will:
bump the npm version, do a git tag (built-in feature of npm version) and pour you a beer ;-)

Publish will only work on an clean git working directory, so all work in progress must have been committed before.

I can't fully test publish myself as I don't have access to the npm publish command, so you'll have to try this yourself next time when you publish (on paper it should work)

I have used a second karma config, as a coverage test on a minified js is not really meaningful and I did not have another way to tell karma to skip the coverage test.

Still todo is the activition of lint as a pretest condition.
Reason for this is that the current Lokijs does not pass jshint due to redefinition of variables.
There are a few ways to fix this:
- rename i to i1, i2 etc
- rename i to j,k,l etc
- isolate these parts in separate functions to restrict scope 

Happy to fix this as well if you indicate your preference.

### DIFF
--- a/karma.build.conf.js
+++ b/karma.build.conf.js
@@ -1,0 +1,73 @@
+// Karma configuration
+// Generated on Wed Mar 25 2015 21:28:22 GMT+0000 (GMT)
+
+module.exports = function (config) {
+  config.set({
+
+    // base path that will be used to resolve all patterns (eg. files, exclude)
+    basePath: '',
+
+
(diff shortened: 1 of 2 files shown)


techfort:
awesome work as usual. Should be able to test it tonight.

PR #138, opened 2015-04-10
Fixed last jshint warnings and enabled jshint as mandatory precondition to npm test

seriousme:
Was less hassle than I initially thought ;-)


techfort:
I'll run it with an empty tankard, i'd better see that beer filling it!

seriousme:
Just make sure you watch the console carefully when you run 'npm run publish' ;-)

PR #143, opened 2015-04-15
fixed onsuccess callback

mahadazad:
Due to jshint fixes few days ago. You made a strict check of `res === null` which fails because res is actually undefined and hence `res === undefined` check fixes the issue.


techfort:
Thanks a lot for this!

PR #153, opened 2015-05-22
Create CONTRIBUTING.md

tomByrer:
Makes a nice link in GitHub webpage UI to this file when a new issue is created.  Also prevents mavericks from messing with your guidelines (WiKi editing is open to all).

Reasoning: Want to add mention if code must be ES5.1 complaint, or what ES6 items you're OK with.


techfort:
Cheers Tom!

PR #155, opened 2015-05-30
find by example

retorquere:


techfort:
Hey @retorquere , linting fails because of a missing semicolon at line 2696. Could you fix that so I can merge. Thank you.

retorquere:
Which linter do you use?

techfort:
jshint

retorquere:
OK, done, and passes jshint

retorquere:
Awesome. What's your release schedule? I mean, when could I expect to find this in the minified browser version?

techfort:
1.3 is due in a few days, let me create tests for your feature and all
going well it will make it in - no problem!

retorquere:
Hey @techfort, is there a release plan for 1.3? I could really use find by example. Using a monkey-patched version of LokiJS right now, but I'd rather not.

retorquere:
I'm passing this as query to `find`: `{"$and":[{"citekey":"grice:defense-of-dogma"},{"libraryID":null}]}`. That is a valid query, right? One of my users is getting an error `TypeError: fun is not a function` at lokijs.js:1:20886 (this is the minified version. can I just use the version from the source directory? Easier to debug)

techfort:
yep - you can use the source directory version. As for 1.3, working on it
at the moment! Bottom line is, any day now, i just have to test out unique
index and update docs.

retorquere:
Cool. The above query looks OK to you? I'll get more debug info soon from the user, as I can't yet reproduce locally.

techfort:
sorry - forgot to comment on the query. Yes, looks ok to me.

retorquere:
I must be missing something WRT to that query -- https://gist.github.com/d4b584541e83a507d319 bombs out

obeliskos:
I was able to reproduce that error and i think the commit i just made would fix it.  Let me know if that fixes for you.

retorquere:
unfortunately it doesn't: https://gist.github.com/950f4445a5ed43819321.

retorquere:
It is definitely related to searching for nulls... I've submitted a separate issue for this, but this one restriction is somewhat problematic for me; I'll have to consider reworking my code to structurally eschew nulls.

This is where the pre-X methods being allowed to change their objects would really help BTW.

obeliskos:
Your example works for me.  I don't know exactly what your shell script is expecting but you should use the final line of code to assign to a results variable... would seem your code is ignoring the result?

retorquere:
The script did ignore the results. In the script I was solely interested in
not getting an error :) again apologies for the panic (which I totally did)
and thanks for the quick fix.

PR #179, opened 2015-06-26
Safari IndexedDBAdapter Issue

kberryman:
The first time you save using indexedAdapter in Safari. you get the following error "TypeError: null is not an object [evaluating 'result.id']".  This is because Safari returns a null result if an index is not found not undefined like Chrome.  I fixed the code to check for both null or undefined.


obeliskos:
Thanks, I had seen that (or something similar to that) before on ios but had not looked into it yet.

PR #186, opened 2015-07-06
Fix reference to global loki variable

morsdyce:
The global loki variable name changed from lokijs to loki thus breaking the angular module.


techfort:
thanks for catching this!

morsdyce:
Sure, will appreciate if you could publish a minor release for this when you can as this breaks angular apps using the latest version.

techfort:
@morsdyce v1.3.7 is published

morsdyce:
Wow, thanks!

Issue #187, opened 2015-07-07
The test described in 'should retrieve records with by()', concerning the 'insert' uses an 'update' instead


tnarik:
Hi,

I was looking forward the 1.3.x addition of 'unique' fields but it seems not to apply to 'insert'. When using inserting a record with a duplicated index, it fails.

If the instead of 'indices' the property 'unique' is used to create the collection (in 'addCollection') a warning is displayed but the entry is created nonetheless.

tnarik:
Closed via pull request #191.

PR #188, opened 2015-07-07
Pull request with test for issue #187 (the issue itself is fixed by #191)

tnarik:
#187

### DIFF
--- a/spec/generic/unique.spec.js
+++ b/spec/generic/unique.spec.js
@@ -26,6 +26,14 @@ describe('Constraints', function () {
     expect(function () {
       coll.update(joe)
     }).toThrow(new Error('Duplicate key for property username: ' + joe.username));
+
+    expect(function () {
+      coll.insert({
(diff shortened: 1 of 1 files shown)


techfort:
@tnarik thanks for the PR! unfortunately it's failing the build because of a failed test in unique.spec.js. Could you check and revise?

techfort:
@tnarik sorry - it looks like your PR is more of a demonstration of the error than a fix. Will work on this asap as it seems to be a major bug.

tnarik:
Hi. I might not have been clear. 
The PR if adding the use case that fails so that it is reproducible, not the fix.
I am not sure if the intention for 'unique' indices is to prevent the insertion as well or not (I think so, but maybe is not), and nodejs is not my strength, but I'll try to take a look at it and provide a fix as well as the updated spec.

tlkiong:
Thank you. I was experiencing this as well.

Issue #189, opened 2015-07-08
Module is undefined in lokijs-indexed-adapter


tlkiong:
568     if(module && module.exports) {
      569       module.exports = IndexedAdapter; 
      570     }

The above code is screaming the error. I am building a cordova app. This error came about when I upgraded LokiJS from 1.3.4 to the current version of 1.3.7. (It broke my app)
V1.3.4 didn't have these lines.

techfort:
Thank you very much for reporting this. Actually @tlkiong if i can be so bold as to take advantage of your generosity, would you have a PR for this which supports `module.exports` but doesn't throw the `module is undefined` error? it will be a couple of days before i can tackle this properly otherwise as i'm up the walls with other projects

obeliskos:
Probably just a simple typeof check would do.  Something like : 
```javascript
if (typeof module !== 'undefined' && typeof module.exports !== 'undefined') {
  module.exports = IndexedAdapter;
}
```
If no one else does, I can change that later tonight.

techfort:
hey @obeliskos - problem with this kind of check is that the second condition will always be checked, and if `module` is undefined then `module.exports` throws an error because it tries to retrieve a property of an undefined. 
So either we do the same trick as in lokijs.js 
```
(function (root, factory) {
  if (typeof define === 'function' && define.amd) {
    // AMD
    define([], factory);
  } else if (typeof exports === 'object') {
    // CommonJS
    module.exports = factory();
  } else {
    // Browser globals
    root.loki = factory();
  }
}(this, function () {

  return (function () {
  // ... etc.
```
or we do nested IFs (if module then if module.exports).

obeliskos:
I think it would use short-circuit Boolean evaluation.  Seemed to work for me in a browser...

Maybe others can test or suggest.

techfort:
@obeliskos you know, the problem is that i really need to give some TLC to
LokiJS in Corodva. There are more and more requests for a ready-made fs
adapter for cordova and this breaks in Cordova (if i get this right).
Only today i discovered IndexedDB is not supported in Cordova for iOS so
it's either localStorage or fs, and most people wouldn't feel very
comfortable with localStorage

obeliskos:
Ok, I'll let you figure out how you want to handle cordova.  This does seem to fix normal browser use in non-commonjs environments though.

techfort:
cool - go ahead and commit that so because having only cordova broken is
better than everything broken :)

[...]

PR #191, opened 2015-07-08
Fix issue #187

tlkiong:
**(1) This is the problematic code**
        3098     // add the object
        3099     this.data.push(obj);
        3100
        3101     var self = this;
        3102     Object.keys(this.constraints.unique).forEach(function (key) {
        3103       self.constraints.unique[key].set(obj);                                        <= error thrown here
        3104     });

**(2) This is the working code**
        3098     var self = this;
        3099     Object.keys(this.constraints.unique).forEach(function (key) {
        3100       self.constraints.unique[key].set(obj);                                        <= error thrown here
        3101     });
        3102     
        3103     // add the object
        3104     this.data.push(obj);

This is because as @tnarik mentioned, it print to console the following "warning" which is in fact an error thrown.
>Duplicate key for property...

In (1), the object has already been added to the collection then only the error is thrown (under set function) - thus, having (2) will solve the problem of "Insert although same unique field"

### DIFF
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -3083,13 +3083,14 @@
         obj.$loki = this.maxId;
         obj.meta.version = 0;
 
-        // add the object
-        this.data.push(obj);
-
         var self = this;
         Object.keys(this.constraints.unique).forEach(function (key) {
+          // Function set will throw error when unique constraint is not honoured
           self.constraints.unique[key].set(obj);
         });
+
+        // add the object
+        this.data.push(obj);
 
         // now that we can efficiently determine the data[] position of newly added document,
         // submit it for all registered DynamicViews to evaluate for inclusion/exclusion


tlkiong:
That was fast. Thank you @techfort

techfort:
@tlkiong @tnarik thanks a lot for your help on this! I have been extremely busy and LokiJS has paid the price of this. I see the `data.push` call happening before the unique check, that's an obvious mistake. 
I want to explicitly thank you for this because this was a major bug and it warrants a minor release, once  https://github.com/techfort/LokiJS/issues/189 (@tlkiong this is yours too ;D )  is fixed.

tnarik:
Cool, and thanks a lot for the great work, I love LokiJS. Do you think it would be good merging the test case as well?

techfort:
I'll check immediately!

tlkiong:
@tnarik thanks a lot for the test case (I was just about to write one then realize you had done it)
@techfort, don't mention it. Really love what you have done here :+1:

PR #200, opened 2015-07-30
Update loadJSON to avoid Unexpected enf of input

Alex-Werner:
In some cases (Error from our node), the database file got erased, only a blank file remains. 
So loadJSON serialize from an empty file throwing this SyntaxError.

### DIFF
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -617,6 +617,7 @@
      */
     Loki.prototype.loadJSON = function (serializedDb, options) {
 
+      if(serializedDb.length===0) serializedDb=JSON.stringify({});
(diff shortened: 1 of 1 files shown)


Alex-Werner:
The database erased happens in these case : 
     1. I use pm2 with watch on, when I change file, it restart, sometimes it's got this problem. 
     2. Sometimes, on full stack size excedeed, node restart, creating sometime that's problem. 
My tought is that while i'm having a insert every 1second (financial data), with an autosaveInterval: 60000. I think that restart node while saving making this problem. I will see if I got more information but it could be interesting to see why.

PR #202, opened 2015-07-31
Adding an options autosaveCallback

Alex-Werner:
It could be handfull to have a callback every time our db autosave (for heapdump, for logging, for emit etc...). 
Exemple :   return new this.loki(dbPath, {
                        autosave: true,
                        autosaveInterval: 100,
                        autosaveCallback:function(){
                            console.log('[MIB] Saved');
                        },

It works with or without the callback (save way than autoloadSave.)

### DIFF
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -497,7 +497,12 @@
         if (this.options.hasOwnProperty('autosave') && this.options.autosave) {
           this.autosaveDisable();
           this.autosave = true;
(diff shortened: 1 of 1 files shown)


obeliskos:
Looks good.

It is consistent with how we are handling autoloadCallbacks so I'm good with it.

If needed, we can modify this later to emit an event for this since it is a repeating event/callback.

PR #206, opened 2015-08-07
improved contains

KimSchneider:
contains returns only results that match string, object or array.


techfort:
all is well what ends well!

Issue #211, opened 2015-08-31
Loki IndexedDb Adapter: Methods that use a LokiCatalog can throw intermittent errors


victormacae:
IndexedAdapter has some methods that needs a LokiCatalog instance, for example, saveDatabase method. This method verify if "this.catalog" has a reference to an existent instance of LokiCatalog, if had, then, the method "setAppKey" is called directly. This method try to access a instance of IDBDatabase, and sometimes "onsuccess" event of indexedDb.open was not be raised yet, consequently, the line " var transaction = this.db.transaction(['LokiAKV'], 'readwrite');" of method "setAppKey" throws a null reference exception because "this.db" is null and cant access "transaction" method.

If the validation "if (this.catalog === null)" was "if (this.catalog === null || this.catalog.db === null)" then forces to create a new LokiCatalog instance, and then, always "this.db" have a IDBDatabase instance.

it's a possible solution ?

Ps.: Sorry about my english. :)

techfort:
hey @victormacae - that makes sense. If you want to give a go to a PR and verify your changes with tests, feel free!

victormacae:
Ok @techfort, i'll do that!

victormacae:
Hey @techfort , do you have an indexeddb mock that can i use in my integration test ? Because when i run the tests debugging in chrome it works, but, in phantomjs "window.indexedDB" dont exists. Or can you giveme another idea ? Thanks!

obeliskos:
Considering I don't have any unit tests myself for indexeddb, you can just check in your changes.  

So the environment you were experiencing intermittent errors was chrome/desktop?

indexeddb mock sounds like a great idea for future, i'll keep my eyes open or if you find one feel free to contribute :) 

Thanks

victormacae:
@obeliskos, The environment that i use and test was chrome, but i belive any usage of indexedDb adapter could throw the same intermittent erros.

Ok, i'll check in my changes and keep my eyes open too looking for a good implementation of a indexedDb mock to contribute with more tests, all right ?

Thanks!

obeliskos:
Checkin looks good, thanks.

Maybe tomorrow I will update our examples/sandbox page to include the indexed adapter script which might be a simple way to pass gists to reproduce future errors.

Feel free to more submit more pull requests if you find anything else.

victormacae:
@obeliskos , very good idea!

Thanks!

PR #213, opened 2015-09-01
Simple fix solution

victormacae:
+ Add validation to avoid intermittent null reference exceptions when LokiCatalog methods trying to access "transaction" function of "this.db" object.

Refers issue #211


techfort:
thanks @obeliskos  and @victormacae !

victormacae:
thanks a lot!

PR #216, opened 2015-09-07
Added a loaded event to the main database class

KodeAndGame:
If this isn't helpful, I'd love to know how I can accomplish what I'm looking for. This allows me to use the Loki constructor with autoload and use an event listener to populate any collections as appropriate. This is especially helpful when the creation of the database and the loading of the collection might be separated for some reason.

```
db = new loki(dbFile, 
	{
		autoload: true,
		autosave: true, 
		autosaveInterval: 10000,
	});

db.on('loaded', function() {
	var reactions = db.getCollection('reactions');
});
```

### DIFF
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -376,6 +376,7 @@
 
       this.events = {
         'init': [],
+        'loaded': [],
         'flushChanges': [],
         'close': [],
         'changes': [],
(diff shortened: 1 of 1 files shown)


obeliskos:
While we already have an option to pass and **autoloadCallback** function within the loki constructor, this style may also be useful.  Thanks!

techfort:
cheers @KodeAndGame  and @obeliskos  for merging.

PR #224, opened 2015-09-18
Calling Database#close callback after saveDatabase completion

Cortys:
Maybe I missed something but I couldn’t find a way to elegantly save a persistent database before exiting because the close callback was sometimes called before the async fs operation completed.

### DIFF
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -731,7 +731,8 @@
       if (this.autosave) {
         this.autosaveDisable();
         if (this.autosaveDirty()) {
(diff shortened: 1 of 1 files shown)


obeliskos:
Looks ok to me.   i'll let @techfort look over and merge since I kind of hijacked his close event to do autosave and this would suppress the emit.

techfort:
@obeliskos @Cortys I had a look and all seems good, going ahead. Holding off the release until date-specific comparison operators are implemented.

PR #225, opened 2015-09-19
Make sure Resultset#filteredrows is an array of numbers

Cortys:
I came upon another small problem...

The evaluate and remove methods of DynamicView check for the position of the given `$loki` id in its Resultset via `indexOf`, which uses `===` internally.

The `filteredrows` in Resultset are in some cases generated via `Object.keys`, which returns an Array of strings.

Since `typeof doc.$loki === "number"` remove and evaluate calls from DynamicView would fail, resulting in an inconsistent DB state where a document could for example be removed from a collection but still be referenced in a DynamicView of that collection.

This issue caused problems for me when loading a persistent collection with persistent DynamicViews from fs and then removing a document. When iterating over the DynamicView there either was an undefined entry or a duplicate of another document in it.

I have not written a test case for this yet since I wasn't sure how your spec is structured.

### DIFF
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -1069,7 +1069,7 @@
     Resultset.prototype.limit = function (qty) {
       // if this is chained resultset with no filters applied, we need to populate filteredrows first
       if (this.searchIsChained && !this.filterInitialized && this.filteredrows.length === 0) {
-        this.filteredrows = Object.keys(this.collection.data);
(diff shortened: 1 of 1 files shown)


obeliskos:
This makes sense, our tests are in the spec folder... would probably go in spec/generic/test.spec.js.  or we could add new 'resultset' specific test spec (use the simpler transforms spec as a template).  not hard to figure out.   

Thanks

Issue #231, opened 2015-10-02
Utilize Object.observe for automatic dirty checks


Cortys:
It would be quite nice if Loki supported automatic dirty checks via `Object.observe` (if available), which would make it unnecessary to call `update()` manually.

Ideally an implementation of this feature could also check whether an indexed property was changed and unlike `update()` only mark the affected indexes as dirty.

This would make code easier and possibly even faster (since no unnecessary index rebuilds will be triggered).

obeliskos:
I suppose feature detection is just checking if that method exists?  Quick search leads me to think maybe this is supported as of node.js version 12.0 (regular -and- harmony)? I suppose all objects could register the same callback and we could dig out the $loki id when it fires.  Would probably need throttling too.

Not sure what @techfort's thoughts on this are, so we'll see what he thinks.  Is this something you would be interested in contributing or just curious?

Cortys:
Sure. I'd like to help with this feature. I would have already started sketching it out if I were more familiar with the codebase. Would you simply add the observers in `loadJSON` and `insert` or are there other ways for an object to get into Loki?

**Regarding Object.observe support:** Yes. As far as I know Object.observe is enabled without runtime flags in Node 0.12+, Chrome 36+ and Opera 30+.

techfort:
@Cortys the idea is nice, however in the future i'd like object clones to be returned from insert/update/read operations, which means you'd be monitoring the wrong object with `observe()`. Or do I misunderstand what you're saying?

Cortys:
@techfort You are right. When working with clones this feature would not be useful.

Now that you mention it, how would clones and the object inflation feature play together?

techfort:
@Cortys the current default inflation pattern is already a clone of sorts, because inflating objects from serialized json effectively creates a new object from a constructor function then copies properties from the source object to the new destination object. So this would not change. If anything, leveraging ES2015 will make it easier to inflate json into a prototype by doing `Object.assign( (new MyObject()), srcObject)`

How does it sound?

Cortys:
@techfort Right, that's when creating the objects. When reading however, do I get a newly inflated instance for each query or do I always get the same inflated object?

techfort:
@Cortys I suppose this would really be optional. If the user needs a full prototype copy at each read he/she should be able to specify so. Probably by specifying a `clone` function? I mean it's all very much optional so it can go to any level of complexity.

Cortys:
[...]

PR #233, opened 2015-10-04
Automatic document updates via Object.observe

Cortys:
As discussed in #231.

Because Object.observe is async, updates that cause a unique key violation fail without throwing and only emit an error event, that cannot be safely associated to the invalid object.

I'm not sure about how much of an issue this is and if it has to be resolved before merging.


obeliskos:
Looks good so far.  Code looks clean, tests pass, and benchmarks are at par with previous (at least without it enabled in our benchmark).  It will take time to look over and i'll let @techfort have plenty of time to review, but looks fine to me.  Maybe a wiki page or paragraph to explain this once it is in main branch?

obeliskos:
Code looks great, I assume @techfort is busy on other tasks so i'll just go ahead and merge this.

If you feel inclined to do so, could you write up a description of how to use this for wiki?

Thanks!

Cortys:
Sure, I'll try to get it done by the end of this week.

Do you plan to release this in 1.3.11 or 1.4?

obeliskos:
@Techfort does all the releases and versioning so maybe he can give you an idea on this... whenever he gets back.  He may need congratulations and a cigar as well ;)

obeliskos:
I do very much appreciate the unit tests by the way... that helps out a lot.

PR #236, opened 2015-10-07
Fix minor bug in autoupdate implementation

Cortys:
I just noticed, that I got the Set fallback implementation wrong.

Nothing severe, since only Chrome 36 + 37 supports Object.observe and does not offer Set.


obeliskos:
NP, Thanks

PR #237, opened 2015-10-08
New service in loki-angular.js

helzgate:
Created a new service called Lokiwork that allows for a more simple angular implementation.  I didn't touch anything else in the original Loki code.


helzgate:
fixing some errors so closing

PR #238, opened 2015-10-08
New service in loki-angular.js

helzgate:
I added an additional service that does a lot of the heavy lifting for the developer in angular, for example, allowing the user to easily load a database from a json file and for the database to automatically re-create if it gets deleted (unless the json file is removed).  The original Lokijs code isn't touched, all i did was add a service called Lokiwork within the loki-angular.js file.  All of my additions are within that service.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ This version of Lokijs for Angular simplifies things to the most basic level bec
 `angular.module('app',['lokijs']);`
 
 ###Configure database template:
-
+I might call this file - json_locations.js
 ````
 (function(){
 	angular.module('app').constant(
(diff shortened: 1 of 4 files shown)


obeliskos:
Hey, this looks really interesting, thanks!  Normally, @techfort would review this since he wrote the original but I believe he may be offline for a few days or a week or so.

I did notice it looked like you had made changes to the several files which may be specific to your fork. 

If you submit a pull request which only includes the **loki-angular.js** file I'll go ahead and merge that since it is not critical path, but some of the changes to the other files do not look appropriate to merge with main branch.

Thanks!

helzgate:
Closing request because i should have only submitted one file.

PR #241, opened 2015-10-09
Improved error reporting and fixed a null id.

helzgate:
Improved error reporting and fixed a null id.  Also, I enhanced the [plunker](http://plnkr.co/3H1kgFKWsr341zsWLZvp) to show off all functions now.


obeliskos:
Thanks

PR #246, opened 2015-10-14
Correct a misprint in DynamicView.prototype.applySortCriteria

VladimirTechMan:
Currently, due to an existing misprint in the property name, the criteria passed as a parameter to the function is not used for the actual sorting.


obeliskos:
Wow... thanks!

PR #251, opened 2015-10-17
Removed some annoying requirements

helzgate:
No longer need to specify a specific collection or db key name.  Multiple collections can now be added in one update or addition.


obeliskos:
Thanks!

PR #254, opened 2015-10-20
$containsNone

KimSchneider:
Didn't find a $not function and required a way to find only entries that don't contain something.


obeliskos:
Thanks!

PR #257, opened 2015-10-24
 Ensure that the data returned by ResultSet never expose a ref. the original collection data

VladimirTechMan:
In the current implementation, when the "persistent" option is configured for a dynamic view, the `evaluateDocument(...)` and `removeDocument(...)` methods of the view modify its `resultdata` to keep them consistent with the changes in the original collection data (after applying the configured filtering, if any).

In the case when there is no actual filter applied to the view, the `resultset.data()` method simply returns a reference to the actual data owned and controlled by the linked collection. (Doing so _is_ actually good for performance, thus I do not think we would need to change that aspect.) Unfortunately, in this case, the current implementation of the dynamic view simply keeps modifying "its" persistent data – in reality, it is making changes to the actual collection data. That causes some very subtle issues and "strange" errors when using dynamic views without a filter (with or without a sort parameter).

The suggested fix ensures that any modifications to a dynamic view's persistent data are only done on a copy of the collection data, fully owned and controlled by that dynamic view. Never on the original collection data.

### DIFF
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -1681,7 +1681,7 @@
         }
         // not chained, so return collection data array
         else {
-          return this.collection.data;
+          return this.collection.data.slice();
         }
       }
 
@@ -2029,7 +2029,7 @@
       // if this is chained resultset with no filters applied, just return collection.data
       if (this.searchIsChained && !this.filterInitialized) {
         if (this.filteredrows.length === 0) {
-          return this.collection.data;
+          return this.collection.data.slice();
         } else {
           // filteredrows must have been set manually, so use it
           this.filterInitialized = true;
@@ -3441,9 +3441,7 @@
     };
 
     Collection.prototype.removeDataOnly = function () {
-      this.removeWhere(function (obj) {
-        return true;
-      });
+      this.remove(this.data.slice());
     };
 
     /**


obeliskos:
Good find and you raise great issues.   Now that I do testing I see that if someone does the following : 

```javascript
var results = coll.find();  // or coll.chain().data();
results.insert({ a: 111, b: 222 });
```
A new document is unofficially inserted into the collection data (without meta) or view updates.  I don't really want that behavior at least by default.  

I am thinking it might be better to fix this issue at the resultset.find() and resultset.data() origination, by making new array copy with whatever is fastest (slice?).  As you point out this may be a (tiny?) bit slower when fetching all documents with blank or no filters.

I am preparing to rework our (optional) collection 'clone' option to provide a more complete implementation of immutable collection behavior.  Probably beyond your concern I just mention it because this fits into that.

If you want to rework this to slice data in resultset.find() and resultset.data() that can work to fix the problem you point out.  Otherwise I can do this later.  Let me know if you have any thoughts on this.  If performance for your usage is impacted a lot we can look at what options we could have for internal array referencing.

VladimirTechMan:
Totally agree with you suggestion. That is actually what I did internally, but I try to be careful with kind of modifications to the original code where that may "clash" with the author's "point of balance" between the code-safety and performance. I am glad that we are on the same page here! And I am looking forward to the addition of the immutable data sets to Loki, which – I agree – may alleviate the concerns around a small performance degradation related to making a shallow copy of the collection data array. (Time will tell.)

To note, I also made an extra small optimization to `Collection.prototype.removeDataOnly`, following up on your recent changes to `removeWhere`). I hope you don't mind.

obeliskos:
Looks great, thanks a lot for pointing this out!

PR #259, opened 2015-10-25
Persistent dynamic view with active sortPriority and no filter re-sorts ready document array

VladimirTechMan:
In the current implementation, the `DynamicView.prototype.data()` method will always call its `performSortPhase()` method when the filter for `resultset` is not initialized (that is, when flag `resultset.filterInitialized` is equal to `false`). That works correctly in all the cases. But in the case where the dynamic view is persistent and the "sortPriority" options is set to "active", that will simply cause the already sorted document array in `resultdata` to be sorted one more time. Thus, degrading the performance without a need.

This fix adds an extra check to detect this special case and skip the sort phase for it. If any question or concerns on it, please, let me know.

### DIFF
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -1681,7 +1681,7 @@
         }
         // not chained, so return collection data array
         else {
-          return this.collection.data;
+          return this.collection.data.slice();
         }
       }
 
@@ -2029,7 +2029,7 @@
       // if this is chained resultset with no filters applied, just return collection.data
       if (this.searchIsChained && !this.filterInitialized) {
         if (this.filteredrows.length === 0) {
-          return this.collection.data;
+          return this.collection.data.slice();
         } else {
           // filteredrows must have been set manually, so use it
           this.filterInitialized = true;
@@ -3441,9 +3441,7 @@
     };
 
     Collection.prototype.removeDataOnly = function () {
-      this.removeWhere(function (obj) {
-        return true;
-      });
+      this.remove(this.data.slice());
     };
 
     /**


obeliskos:
Your pull requests are simultaneously very simple and very complicated :) 

Just curious what is the use case for this, you want the whole collection sorted for ultra optimal read speeds?  Do you plan on adding filters later or is this just for optimal (entire) collection sorting?

If you did an applyFind() with no parameters it should be the same as 'find all' similarly to collection.find() with no parameters or collection.chain().data() will return all documents.

With an empty applyFind() you have technically set the filterInitialized flag on its internal resultset so I think it may not perform sort phase until docs are added/updated/removed.

That Boolean logic is a little complicated to understand so if we can establish a protocol for your use case I would rather properly 'initialize' to your use case.  Let me know if your use case is other than what I guessed.

VladimirTechMan:
Agree on the somewhat complicated logic of that condition. :)

As for the use case: you guessed it right. It is an app to examine logs from local and remote applications, both from files and streamed in near-real-time. Thus, the database is built dynamically and changes (by appending more records) while the users interacts with the app. The app, ideally, needs to be fast and responsive to both user inputs (changing the sort criteria and applying filters, "on the fly") and new records arriving. I checked several options and Loki looked good as a staring point. It does not fulfill all my needs at the moment – but I am okay to extend it myself. Where I see some issues (or potential issues) with the current implementation, I may bother you with pull requests periodically -- those that I consider as general ones, not on the app-specific ones that I am doing.

Where the changes proposed do not look like something you'd like to see in the "general-purpose" Loki code, sure thing, just tell me so. That is fine. On the above one, I considered it as a use case potentially applicable to some generic scenarios for other users of Loki (though a fraction of those scenarios would be small, I agree), thus I bothered you with it.

I may still offer a few more code changes going forward, if you don't mind.

obeliskos:
Yea, no problem at all... in fact its good to hear ways people might be using this that I didn't think about. 

We might do any of these things : 
- Add some kind of clear all filters where you can completely clean and reapply new filters.
- Optionally allow naming filters so they can be removed independently

The whole filterInitialized is adding unneeded complexity i want to get rid of, possibly : 
- Initialize the filters to a single empty filter
- when they add a real filter we can clear the empty filter before adding a real one. 
- repurpose filterinitialized to be a filterEmpty flag or just get rid of it and use named 'empty' filter which only affects applyFind and applyWhere.  No more filterInitialized in the more complicated queue and perform sort phases.

Yea hope I am not frustrating you... just trying to clean up my code as well since this is most complicated layering of several classes and I see even the existing logic as less than clear in this situation.  

I like the idea of more dynamic pipelines, adding filters is simple since my applyFind and applyWhere just do incremental resultset find and where ops.  Removing filters would need to (behind the scenes) create a new resultset and reapply filters since these chained ops can't be undone to the resultset. 

So this is rough guess as to possible approach (I think I will probably want to help out with this) : 
- make filters optionally named (add second 'name' param to applyfind() and applywhere()) and store in the filterPipeline objects
- applyFind and applyWhere will remove any filters named 'empty'
- maybe leave filterInitialized flag but add queueInitialization phase, if no filters were added by user after they yield thread we will add 'empty' filter and apply it to resultset.
- get rid of filterinitialized checks in anything other than applyfind, applywhere, and queueInitialization()
- a new clearFilters() method may clear filterPipeline and queueInitialization.
- a new removeFilter() method may clear filterPipeline and either run filters or have queueInitialization do it.

I think this may split up the 'dynamic' filters initialization / reinitialization logic from updating, sorting, evaluation and retrieval logic.

That might take a while... maybe a week for my off hours dev time, if your time constrained perhaps you can work off your fork in the meantime.  If you feel comfortable enough with what i suggested perhaps we can collaborate towards this.

VladimirTechMan:
Thanks. Our "lines of thought" look well-aligned, to me, – what you said is very close to what I was thinking myself in relation that code, in general, while investigating it over the past days/weeks. And also when giving it an extra thought today, after your first reply. Introducing an "empty" filter (applied by default when there is no real filter set) and re-factoring / "re-purposing" the "filterInitialized" flag – or removing a need in that one altogether – sounds as a good plan to me.

(As you might tell from my initial changes above, like the previous fix from me, I tried to "fit" it into the current logic, not changing the underlying concepts dramatically. But as long as you are considering a more radical improvement, which I agree with, – that does not frustrate me, no worries. And, even more, I am glad to know it.

I am trying to implement something like what you described in my code already. A lighter version. For my current purposes, the actual "pipeline" inside a dynamic view would always be either just no filter (that is, the "empty filter" in terms of what we discuss now) + a specific sort criteria. Or it would be one "real" filter  [which in my case is always (purposely) using `applyFind()`, leveraging the internal optimizations based on pre-configured column indexes]  + again, a specific sort criteria. Thus, for my app, there is always a sort part at the end of the pipeline. And it is preceded either by no filter ("empty" filter) or a single "Mongo-style" filter. That all should not make your plans on the more extended logic, with the named individual parts of the filtering chain, any less useful in general, I think – thus, please, keep considering it in your road-map for Loki's evolution.

If I may do a suggestion here, then it would be to take that in two "phases" maybe: Introduce the default "empty" filter and possibly re-purpose or remove the "filterInitialized" flag during that process. Plus, add a capability to `clearFilters()`  (all filters at once, it should be easy). Do that as the initial "phase". And then, as the "second phase", extend that with the named individual filtering parts, allowing to add/remove them separately. _Does that makes sense?_

Though my load during the next few weeks may be changing quite a bit. But, in general, I definitely do not mind cooperating more on making those changes. That is in my own interest too.

BTW, It may also be good to exchange the e-mail contacts and/or the Skype / Hangout / "other communication app" contacts going forward. May simplify coordination. Just an option.

Let me know what you think.

obeliskos:
Ok, yea maybe at first I will check filterPipeline in data() and (if its empty) I will add empty find filter synchronously and remove all checks for filterinitialized, until a proper initialization phase can be added.  The 'getting rid of empty filter' is more important to making sure that subsequents inserts and updates can evaluate as fast as possible.  For 'first draft' of this you could force clearFilters() before setting first (or changing) filtering if you do not plan on adding filters before data() is called.

Will hopefully start looking at this late today.

VladimirTechMan:
Thanks. That sounds like a good plan. I think it makes sense if I let you do the first draft that you want to share with me, then catch up on that – try it and collaborate with you.

Just in case, on my contacts, so you know and can reach out directly for quick questions and alike: The same login name as here, at GitHub, is also my e-mail at gmail.com (and thus, also, my account in Google Hangouts) and my contact name in Skype. Feel free to reach out (just, on your first contact through those channels, make a quick note for me that it is you, so that I know -- the "default" request messages are not very informative typically).

Looking forward to hear back from you soon.

VladimirTechMan:
With the course of action we agreed upon, I have just closed this request – I don't think we need to keep it open. For discussions and tracking the progress with the planned changes, let's use our private contacts that we exchanged. (And we can create a new pull request or issue if we want to expose a part of the message exchange to the public.)

PR #262, opened 2015-10-29
Make the type of errors thrown by LokiJS more consistent and uniform

VladimirTechMan:
Using Error or TypeError instead of pure string messages. (Just a quick general code "cleanup" thing, as that aspect caught my eye in the code.)

### DIFF
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -1774,7 +1774,7 @@
               }
             }
           } else {
-            throw 'Do not know what you want to do.';
+            throw new Error('Do not know what you want to do.');
           }
           break;
         }
@@ -1978,7 +1978,7 @@
       if ('function' === typeof fun) {
         viewFunction = fun;
       } else {
-        throw 'Argument is not a stored view or a function';
+        throw new TypeError('Argument is not a stored view or a function');
       }
       try {
         // if not a chained query then run directly against data[] and return object []
@@ -2070,7 +2070,7 @@
     Resultset.prototype.update = function (updateFunction) {
 
       if (typeof (updateFunction) !== "function") {
-        throw 'Argument is not a function';
+        throw new TypeError('Argument is not a function');
       }
 
       // if this is chained resultset with no filters applied, we need to populate filteredrows first
@@ -3110,7 +3110,7 @@
       }
 
       if (property === null || property === undefined) {
-        throw 'Attempting to set index without an associated property';
+        throw new Error('Attempting to set index without an associated property');
       }
 
       if (this.binaryIndices.hasOwnProperty(property) && !force) {
@@ -3343,7 +3343,7 @@
 
       // verify object is a properly formed document
       if (!doc.hasOwnProperty('$loki')) {
-        throw 'Trying to update unsynced document. Please save the document first by using insert() or addMany()';
+        throw new Error('Trying to update unsynced document. Please save the document first by using insert() or addMany()');
       }
       try {
         this.startTransaction();
@@ -3401,7 +3401,7 @@
 
       // if parameter isn't object exit with throw
       if ('object' !== typeof obj) {
-        throw 'Object being added needs to be an object';
+        throw new TypeError('Object being added needs to be an object');
       }
       /*
        * try adding object to collection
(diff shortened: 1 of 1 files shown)


obeliskos:
Thanks!

obeliskos:
Hey @VladimirTechMan and @techfort I was wondering if we should invoke emitted event handlers with try/catch to avoid allowing them to disrupt internal processes?

A recent pr #509 is using temp file name during save which is renamed to original on successful save.   This temp file is being generated but not renamed in our unit tests.  I found the test in test.spec.js in the 'stepDynamicViewPersistence' test which raises an exception in the close event, which it calls and verifies is thrown as an expectation.  

So @VladimirTechMan is that functionality you are depending on, and (both of you) should we prevent listeners from throwing uncaught exceptions?  I would think each listener could catch within themselves if they wanted notification.

Any thoughts?

VladimirTechMan:
@obeliskos @techfort _Hi guys. Sorry about the delayed reply. Back from two consecutive business trips – and then hitting the road again tomorrow again, on my employer's business. But catching up with the things on personal projects over this weekend._

**Dave, to your question to me:** No, I do not depend on that specific part and functionality in my code. Nor do I intentionally use thrown exceptions for anything outside reporting errors.

More generally, it is an old discussion (for many programming languages) whether using throw with try / catch beyond error reporting and as a generic control flow statement is useful and good. In most cases, the consensus is that it is not – in particular, due to the fact that handling try / catch in many languages cannot be done very efficiently (some functional programming languages being a kind of exception, though). In the JS code, dealing with try / catch for the modern JIT engines is not easy and cheap either. They usually cannot apply most performance optimizations to the functions containing try / catch in them. Thus, with Loki's focus on performance, by default, catching all the exceptions would possibly be a little bit of concern, to me. Though, on that side, if we would want to add those statement on non-critical paths only, the effect would be unnoticeable performance-wise.

The other obvious downside is that just "swallowing" exceptions from external code inside Loki would, potentially, prevent the developers from noticing and fixing them in their code. Which may result in much more subtle and hard(er) to debug situations for them (and them result in them blaming us for the specific design approach on the exceptions from their code).

VladimirTechMan:
@obeliskos @techfort Thus, overall, my view is, **(1)** that JS developers should reply on throw and try / catch as a mechanism for reporting unhandled errors in the code, not as a universal flow control mechanism. And **(2)** that they should handle exceptions in their own code, when that code may result in exceptions.

The only downside is that doing that in every required place across the code can be rather verbose and tedious, at times. If we are only concerned about our test cases here, then, I guess, we may simply correct it on the test-case level ourselves. If it is a more generic concern from developers, in terms how easy it is to use Loki in their code. Then, again, we may add a few safe-guard try / catch blocks to the non-critical paths. That is one option. And the other (potential) option is to enhance our custom event emitter object prototype so that can take an optional parameter (a flag) and, depending on it, either do just the same handling as now or loop through the added event handlers with the try / catch statement around them – and then possibly call a custom "onerror" handler, if added, inside the catch part. That would be rather a small amount of additional code for us in the library. And minor performance impact for the code not using that flag-enabled logic. The only question is if that is worth of the effort at all – does our users need that extra "sugar"?

_Let me know what you think._

obeliskos:
Yea your probably right, especially when those events can be emitted on object insertion which would be performance sensitive.  I have added the non-renamed file to our git ignore to avoid it being accidentally checked in.  If you are not dependent on that assertThrows for 'Throw error on purpose on close' functionality I may remove it or convert our unit tests to use our new LokiMemoryAdapter and not touch the filesystem.

Thanks for being a sounding board on this, I think you are right that we probably should not try/catch errors thrown in our event listeners.

VladimirTechMan:
@obeliskos I am certainly not dependent on that specific assertThrows part, thus feel free to proceed with the related modifications. BTW, I really like the addition of LokeMemoryAdapter, for testing / debugging purposes and alike. Just looked at that today. Good job, Dave! And thanks a lot for all your help in communicating with and supporting the users of Loki in the recent few months, while @techfort and myself happened to be tied up quite a bit.

PR #263, opened 2015-10-30
Quick fix to method removeTransform() of Collection

VladimirTechMan:
Just another quick correction upon noticing a missing reference to `this` in the code.


obeliskos:
Good catch, Thanks!

PR #270, opened 2015-11-07
Added verbose option to enable/disable logging

### LINKED ISSUE #269: Remove all console logs/warnings
Since Loki is usually included into other applications I think it should not log anything by default so that the developer has full control over the output of his/her application. This is especially relevant when developing CLI tools.

I think disabling all logs and offering a flag to reenable them would be an appropriate solution.

Cortys:
Fixes #269

This adds a `verbose` option the the Loki constructor.

Loki instances and its Collections will only log errors and warnings to console if the verbose flag is set. It is disabled by default.

Apart from that I also changed the IndexedDB adapter to throw instead of log if IndexedDB is not supported. I think this behavior is more reasonable and also makes the adapter completely non-logging. Thus the verbose option of a Loki instance does not have to be propagated to control logging in there.

I left all „fallback“ logs (if no callback is provided) where they are so that using Loki in a REPL is still possible. Since Loki itself always provides a callback to all internal calls those logs should normally never appear.

Apart from that the only logs that are left can be found in the `loki-angular.js` module. I’m not sure how to handle those since they are not directly related to a single Loki instance and thus not related to any specific verbose flag value.


obeliskos:
Looks great, thanks!

PR #283, opened 2015-11-24
Add, update, and remove filters in a view by giving them unique IDs

VladimirTechMan:
Building on top of the recent fixes and improvements in DynamicView, this change adds a second optional parameter to applyFind() and applyWhere(), which acts as the unique ID of the filter being applied for the first time. The ID can be either a unique string or a unique number (a mixture of two types will work fine as well, technically – but from the code clarify perspective, the users are better to stick with just one of the two in their code, on top of LokiJS.)

Re-using a previously used unique ID with applyFind() and applyWhere() will result in replacing the previously added pipeline filter with that ID. To remove a filter by its ID, use removeFilter().

Questions and comments are welcome.

### DIFF
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -2565,59 +2565,146 @@
       return this;
     };
 
+
     /**
-     * applyFind() - Adds a mongo-style query option to the DynamicView filter pipeline
+     * Implementation detail.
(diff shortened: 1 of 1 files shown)


obeliskos:
Looks great, thanks!

techfort:
@VladimirTechMan awesome, thank you! thanks @obeliskos for merging

PR #292, opened 2015-12-06
Pass the proper view reference when emitting the 'rebuild' event in queueRebuildEvent

VladimirTechMan:
Currently the callback function inside `setTimeout()` passes local `this` instead of using `self`.


obeliskos:
Thanks!

PR #313, opened 2016-01-16
Fixing misprint in variable + minor code improvements

VladimirTechMan:
Fixing the initial misprint in the code (`cl = cl.length;`), adding a missing return statement for `result`, and consolidating the variables a little bit (`cd` and `cl` essentially duplicated existing `data` and `len`). Finally, making a small improvement to the copy loop, to avoid repeating checks inside (aiming at bigger data sets).

### DIFF
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -2090,8 +2090,10 @@
      */
     Resultset.prototype.data = function (options) {
       var result = [],
-        cd,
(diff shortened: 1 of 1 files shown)


obeliskos:
Thanks, much cleaner and thanks for correcting the 'misprint' (bug) :)

PR #314, opened 2016-01-17
Code improvement to speed up sorting on larger result-sets

VladimirTechMan:
Changing the code to use a (pre-calculated) direct reference to the data array. While it won't make a change on small-to-medium data sets, it starts paying off really nicely on bigger data sets (*think of 5-10K records and more*), where the number of calls to the criteria function from the native `sort()` becomes just huge (*esp. in the "average" to "worst" cases*).

### DIFF
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -1296,11 +1296,9 @@
 
       var wrappedComparer =
         (function (userComparer, rslt) {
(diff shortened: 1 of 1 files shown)


obeliskos:
Looks great, thanks!

techfort:
Ditto!

PR #317, opened 2016-01-18
TTL daemon issue

moshewe:
This is TTL daemon working with .where(fun) for filtering out aged documents. I've tested this live in production, but please do run more tests!

### DIFF
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -169,12 +169,12 @@
         return a === b;
       },
 
-      $dteq: function(a, b) {
+      $dteq: function (a, b) {
         if (ltHelper(a, b)) {
           return false;
         }
 
-        if (gtHelper(a,b)) {
+        if (gtHelper(a, b)) {
           return false;
         }
 
@@ -225,8 +225,8 @@
         }
 
         checkFn = containsCheckFn(a, b) || function () {
-          return false;
-        };
+            return false;
+          };
 
         return b.reduce(function (prev, curr) {
           if (prev) {
@@ -246,8 +246,8 @@
 
         // return false on check if no check fn is found
         checkFn = containsCheckFn(a, b) || function () {
-          return false;
-        };
+            return false;
+          };
 
         return b.reduce(function (prev, curr) {
           if (!prev) {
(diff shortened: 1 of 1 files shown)


obeliskos:
I'll just make a few observations and let techfort handle this pr : 

- It would be helpful to not have your automatic style changes in the p.r. - makes it difficult to see your changes.
- I noticed your ttl function operates on meta.updated which would not exist on a new object and may yield a NaN in your age calculation, is this considered?
- you have some jshint errors in your p.r., can you run npm run test before submitting p.r. to verify all checks and tests are passed.

moshewe:
* Style: right, sorry about that.
* ttl now works with meta.updated || meta.created
* My bad, I just ran it though our application instead of normal testing. I'll have to set up a separate machine for this as I'm working on windows now :(

obeliskos:
Clicking on details for the failure, you can then click on (for example) the nodejs:4.0 line and view the output to see some missing semicolon jshint errors.

I do my building on another box since my primary machine is windows.  On windows you could do : 
npm install -g jshint
and then 
jshint src/lokijs.js
to see jshint errors

Not sure why the other 'warnings' are suppressed when run as part of our build process, I think @seriousme did some npm magic to hide the warnings which were not relevant or that we could not easily fix.

moshewe:
OK
I'ved fixed the jshint stuff, now I just have to set up another box for the rest.

seriousme:
Its been a long time since I touched the stuff, but I don't recall suppressing warnings on purpose ;-)
I did submit PR  to fix (not suppress ;-))  jshint warnings and a fix to make jshint a mandatory precondition to npm test.( #138)

obeliskos:
ah ok sorry... yes I remember you doing a lot with this back then.

moshewe:
I've just tested it on a linux machine, and all tests are OK.

techfort:
great stuff - can you provide a usage example so i can include in the docs?

moshewe:
```
var collection = new Loki.Collection('collection_name',
    {
        ttl: 1000 * 60 * 5 * 3, //15 mins
        ttlInterval: 100 * 60 * 20 //20 mins
    })
```

moshewe:
The daemon cleans up your collection event every `ttlInterval`, removing all items of age > `ttl`

techfort:
@moshewe awesome stuff!

PR #318, opened 2016-01-18
Speed up DynamicView's evaluateDocument() for newly inserted records

VladimirTechMan:
One more code improvement that helps when handling insertions of new records into large collections that have attached dynamic views. By passing a flag indicating that `evaluateDocument()` is called upon adding a new record, doing linear scanning of the indices in the `filteredrows` array is not necessary. That helps when `filteredrows` grows bigger, but in my tests that brings a notable improvement even on smaller filtered sets in dynamic views.

The patch also makes a few minor improvements to the body of `evaluateDocument()`.


obeliskos:
Makes sense, thanks!

PR #319, opened 2016-01-18
More optimal way of mapping array to array of element indicies

VladimirTechMan:
An easier and quicker way to get an array of elements' indices from the original array of elements. Just returning an element index passed as an argument to the callback function of `Array.prototype.map` works better than creating a temporary array of keys, converting original indices to strings and then back to numbers. Like the other recent commits I sent, this one helps (and is well noticeable) on bigger collections.

### DIFF
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -1171,7 +1171,7 @@
     Resultset.prototype.offset = function (pos) {
       // if this is chained resultset with no filters applied, we need to populate filteredrows first
       if (this.searchIsChained && !this.filterInitialized && this.filteredrows.length === 0) {
-        this.filteredrows = Object.keys(this.collection.data).map(Number);
+        this.filteredrows = this.collection.data.map(mapArrayElementToItsIndex);
       }
 
       var rscopy = this.copy();
@@ -1291,7 +1291,7 @@
     Resultset.prototype.sort = function (comparefun) {
       // if this is chained resultset with no filters applied, just we need to populate filteredrows first
       if (this.searchIsChained && !this.filterInitialized && this.filteredrows.length === 0) {
-        this.filteredrows = Object.keys(this.collection.data).map(Number);
+        this.filteredrows = this.collection.data.map(mapArrayElementToItsIndex);
       }
 
       var wrappedComparer =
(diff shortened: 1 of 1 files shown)


techfort:
hey @VladimirTechMan , as usual streams of gratitude come your way from us. Can you benchmark old code against yours? I am truly satisfied that performance is going to be better, but i lack the time to put a decent test together right now, and i'd love to see numbers do the talking :)

VladimirTechMan:
Hey @techfort . Right at the moment, I don't have a public repo or sandbox to show you (I am sorry) -- that possibly would be the best way. But the changes I am submitting go directly from the project that I am working on (it will be open sourced eventually, that is the intention) and my tests of the browser app using Loki as its inner in-memory DB. The app itself is a kind of log browser for a new log and message capturing standard format (which I am working on) -- to be open-sourced with specs too. I can easily (and proudly :)) ) give you the changes related to my recent commits:
* Loading a file of about 50K JSON-records (parsing it into JSON records, adding those to Loki, showing the resulting log in the UI (table-arranged UI) and some other things.
* The fix submitted yesterday reduced the time of loading and handling all 50K records from the file into the app from over 50 secs to abour 12 secs on my machine.
* The fix submitted earlier today brings it dows to ~10 secs.
* The above change reduces it to ~8 secs.

Those improvements are critical to my purposes (as I need to better understand the limits I can achieve – Loki is the most critical part of it now). I plan to continue working on this part for some more time, which may result in a few more proposed code changes.

In any case, if you feel concerned about the changes made – in general or in relation to the specific goals of my app described above – no worries, just let me know. I am using my own custom version of Loki right now anyway – just bringing back what I feel as good general-purpose improvements. I definitely hope that some day I will be able to get back to installing the basic version of Loki with npm and referencing it in my (to-be) open-source project. Would be just perfect!

techfort:
@VladimirTechMan you know what - that sounds good enough to me :) Loki would love to avail of all the improvements you made!

obeliskos:
Looks great, thanks :)

VladimirTechMan:
Thanks, @techfort. I appreciate the positive response. And kudos to you and @obeliskos (and other contributors to LokiJS) for working on this project. "An in-memory DB developed with speed in mind, as its first priority" – it is one of the critical things for the success (fingers crossed) of my initiative.

I definitely consider contributing back some extra general-purpose APIs that I've developed on top of Loki. Before that I will need to do some more work and iron out some of the major issues and inefficiencies, both in my own code and (a few) in LokiJS – on a few of those, I am already in discussion with @obeliskos.

Just to mention, there is one more general issue that I have in my list, but I suspect that it may be a kind of very specific "edge case" to you. (I will need to keep a fix for that in my own code in any case.) As I am not quite sure how you'll feel about it, let me open it just as an issue for now, not a merge request. Then we can discuss. If you recognize it as a defect – or a specific enhancement – that is worth fixing it in the public version of LokiJS. Then it will be really wonderful to have it amended in the general repository too. Anyway, let me first open that new issue (later today). Then we'll see.

PR #322, opened 2016-01-23
Implement method count() in Resultset and DynamicView

VladimirTechMan:
Adding the `count()` method to `Resultset` and `DynamicView`. That way that the number of represented documents can be obtained easily (and more efficiently then doing something like `data().length`).

To-do: Reflect this part of the API in Loki's public documentation, eventually.

### DIFF
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -2073,6 +2073,18 @@
     };
 
     /**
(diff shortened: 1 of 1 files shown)


obeliskos:
Looks great, thanks.

PR #325, opened 2016-01-25
Enhancing Loki with new operators '$keyin' and '$nkeyin'

VladimirTechMan:
I did not see those operators in MongoDB's query "language". I just did found them very useful in my own practice, as an alternative to '$in' and '$nin' when a bigger set of string values need to be used in the query. Checking if a key exists (and is not equal to `undefined`) in the JS-object is notably more efficient than a repetitive execution of `indexOf()` over a set of values contained in the array (except for very-very small arrays).

*Note: I also made a small change to the comparison JS-operators used for the `$in` and `$nin` Loki-operators to use strict comparisons. For the 'number' type that tends to be a tiny little bit more efficient in modern browsers. (If any concerns on that, let me know.)*

### DIFF
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -206,11 +206,19 @@
       },
 
       $in: function (a, b) {
-        return b.indexOf(a) > -1;
+        return b.indexOf(a) !== -1;
       },
 
       $nin: function (a, b) {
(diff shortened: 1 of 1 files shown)


techfort:
looks great thank you

PR #326, opened 2016-01-26
Safe check for window variable

hampusohlsson:
When building isomorphic apps, node process crashes as `window` is not defined. Using `typeof window !== 'undefined'` solves the issue.


obeliskos:
Thanks!

PR #327, opened 2016-01-27
Call loadDatabase callback with parsing error

guewen:
When the database's JSON cannot be parsed back, we have to call the
callback function with the error.  This is important as the
loadDatabaseCallback can be called asynchronously, and in that case, we
have no means to intercept the error if cFun is not called.

With this fix, the error can be handled in the callback function that is
passed to Loki.loadDatabase().

### DIFF
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -971,9 +971,17 @@
 
         this.persistenceAdapter.loadDatabase(this.filename, function loadDatabaseCallback(dbString) {
           if (typeof (dbString) === 'string') {
-            self.loadJSON(dbString, options || {});
-            cFun(null);
-            self.emit('loaded', 'database ' + self.filename + ' loaded');
+            var parseSuccess = false;
+            try {
+              self.loadJSON(dbString, options || {});
+              parseSuccess = true;
+            } catch (err) {
+              cFun(err);
+            }
+            if (parseSuccess) {
+              cFun(null);
+              self.emit('loaded', 'database ' + self.filename + ' loaded');
+            }
           } else {
             if (typeof (dbString) === "object") {
               cFun(dbString);


guewen:
On a side note: I came across this issue due to another bug I'm still investigating. 
Using https://github.com/cosmith/loki-cordova-fs-adapter on a iOS, I have situations where at some point my database file contains 2 times the db's JSON, so it has 2 json roots and is thus impossible to parse. My quick workaround is to catch the parse error and to delete/create it afresh. My feeling is that my issue comes from cosmith/loki-cordova-fs-adapter but if you have any idea...

```js
    db = new Loki(dbname, options);
    db.loadDatabase(options, load);

    function load(err) {
      if(err) {
        // If for any reason, the database is corrupted, we drop it and create
        // a new one. The data will be synchronized again so it is not a big deal.
        if(db.persistenceAdapter.deleteDatabase) {
          // using LokiCordovaFSAdapter
          db.persistenceAdapter.deleteDatabase();
        } else {
          // When we are in a desktop browser, loki uses localStorage
          // and loki has no 'deleteDatabase()' core method.
          $window.localStorage.removeItem(dbname);
        }
        // creates a fresh database
        db = new Loki(dbname, options);
        db.loadDatabase(options, load);
      }
      readyDefer.resolve();
    }
```

Do you think a `Loki.deleteDatabase()` would be good? I had to manually call the persistenceAdapter's function and remove the key in the localStorage here, would you accept if I propose a PR with `Loki.deleteDatabase()` and their counterparts in the adapters?

techfort:
Hi @guewen - that "doubling" of the JSON is a cordova-fs adapter issue, I had another developer experiencing it, but I was under the impression it was resolved. Also, the `deleteDatabase` option would be great! Thanks for all the work.

guewen:
> that "doubling" of the JSON is a cordova-fs adapter issue, I had another developer experiencing it, but I was under the impression it was resolved.

Thanks for the confirmation (and for the merge as well)! I will check that. 

> Also, the deleteDatabase option would be great!

Noted

perry:
@guewen I couldn't find an issue for the doubling up issue, so FWIW I was also experiencing it when manually calling `db.saveDatabase()` twice in a period of 1-2ms.

guewen:
> @guewen I couldn't find an issue for the doubling up issue, so FWIW I was also experiencing it when manually calling db.saveDatabase() twice in a period of 1-2ms.

I still have this issue, and what you explain was my assumption. I suspect it happens when I close the application, `Loki.close` saves the database, an autosave might be already ongoing few ms before.

PR #329, opened 2016-01-28
Add deleteDatabase on Loki and adapters

guewen:
As proposed in https://github.com/techfort/LokiJS/pull/327


techfort:
looks great thank you

PR #336, opened 2016-02-07
Implement  $size operator (fixes #334)

### LINKED ISSUE #334: Implement $size operator
A `$size` operator would be very useful when trying to match for exact matches in arrays.

For example, let's say that all documents in a collection have a key named `test` and that key holds an array (document example: `{name: 'foo', age: 12, test: [3, 7, 2, 4]}`.

A `$size` operator could be used like this:

    .find({'test': {'$size': 3}})

to find all the documents that have 3 elements in their `test` key.

alexandernst:


src/lokijs.js
```diff
@@ -359,6 +359,14 @@
           }
         }
         return false;
+      },
+
+      $size: function (a, b) {
+        if (!Array.isArray(a)) {
+          return false;
```
alexandernst: Not an expected value. What should be returned? @techfort

src/lokijs.js
```diff
@@ -359,6 +359,14 @@
...
+        } else {
+          return a.length == b;
```
techfort: @alexandernst always use `===`

src/lokijs.js
techfort: `false` is the current value to return, is that what you mean @alexandernst

alexandernst: @techfort I mean, should be something else returned instead of `false`? Or is that ok?

src/lokijs.js
alexandernst: Good point, let me push that change.

alexandernst:
One more thing. I can add docs too, if you want me to do so. If "yes", where should I add that? @techfort

src/lokijs.js
techfort: oh ok - well as I understand the operator `$size` it tests an arbitrary number against the length of an existing array, so this is either true or false. Well in this plane of existence at least. So yeah, we're good.

techfort:
@alexandernst docs as in jsdoc or the main docs in the website? either are more than welcome.

alexandernst:
@techfort In the website. I can't find how/where are those generated. Also, I just realized some of the operators in `master` aren't available in the docs in the website.

techfort:
@alexandernst docs on the website are done manually at the moment. I had plans to migrate to some kind of markup solution of sorts (like readthedocs) but time is seriously lacking in the last 4 months. If you want to contribute to the docs on the website that would be outstanding ... not many devs willing to do that.

PR #349, opened 2016-02-21
Make removeCollection work (fixes #348)

### LINKED ISSUE #348: removeCollection doesn't work
Consider this simple test code:

```
loki = require('lokijs')
var db = new loki('sandbox');
var items = db.addCollection('items', {clone: true});

items.insert({ name: "Foo", age: 42, car: "no", bike: "yes" });

db.removeCollection('items');

console.log(items.find({}));
```

I'd expect `null`, `undefined`, an empty array or pretty much anything `null`-ish to be returned, instead of the `{ name: "Foo", age: 42, car: "no", bike: "yes" }` object it was inserted earlier.

alexandernst:
The reference count isn't decreased just by slicing the collection itself from the collections array. For the reference counter to actually be decreased (and make the GC kick in), each property of the collection we're trying to remove must be null-ified.

An easy way would be to just do

```
this.collections[i].data = null;
this.collections[i].binaryIndices = null;
this.collections[i].flushbuffer = null;
...etc...
```

but by creating a new collection we know that even if a new property of the `Collection` class is added in the future, we'll be safe and we'll remove it (as we're iterating all the proper keys of the class itself).

### DIFF
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -756,6 +756,13 @@
 
       for (i = 0; i < len; i += 1) {
         if (this.collections[i].name === collectionName) {
+          var tmpcol = new Collection(collectionName, {});
+          var curcol = this.collections[i];
+          for (var prop in curcol) {
+            if (curcol.hasOwnProperty(prop) && tmpcol.hasOwnProperty(prop)) {
+                curcol[prop] = tmpcol[prop];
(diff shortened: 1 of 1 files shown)


techfort:
@alexandernst great stuff and good point!

PR #351, opened 2016-02-24
Avoid crash when dbString is null

javiercorsini:
There is a bug in ECMAScript that considers null type to be 'object'. If the loadDatabase of the persistenceAdapter returned null, the app crashed when it tried to get null.collections


kevinprotoss:
+1

obeliskos:
Thanks!

PR #358, opened 2016-02-25
Fix #344

### LINKED ISSUE #344: [Bug] Chained find() and indices stopped working with 2d21fcc9f683c1089522a1781318a045321fb617
I'm testing the `indices` option with chained `find`s and I found a bug. 

Consider this very simple example:

    loki = require('lokijs')
    var db = new loki('sandbox');
    var items = db.addCollection('items', {
        indices: ['tree_visibility']
    });
    items.insert({ tree_visibility: true, order: 0 });

    console.log( items.chain().find({ 'tree_visibility': true }).simplesort("order", false).data() );

The `console.log'` won't return anything at all with `1.3.11`, but will work as expected with `1.3.10`.

More clues:

    console.log( items.chain().find({}).find({ 'tree_visibility': true }).simplesort("order", false).data() );

will work as expected with both versions.

Doing a bisect I found this bug was introduced in 2d21fcc9f683c1089522a1781318a045321fb617

alexandernst:
This commit fixes bug #344 .
The bug happened at 2d21fcc9f683c1089522a1781318a045321fb617 . The problem with that commit is that it didn't had in mind that the props can be actually of a boolean type, which led to very awkward situations in which `ltHelper` would say that `true` is smaller than `false` (which is not true).

This patch fixes this behavior by checking the type of both props, and if both props are `boolean`, then it compares them with the following logic:

| prop1 | prop2 | ltHelper | gtHelper |
|---------|----------|------------|-------------|
| true   | true     | false      | false       |
| true   | false    |  false     | true        |
| false | true      | true       | false       |
| false | false    | false      | false       |

The code could be further shortened, but I leaved it as is on purpose. Reading it at it's current state doesn't require too much thinking, but shortening it further will require further time to read/debug and it won't bring any performance improvements at all.

@obeliskos Are you fine with this patch? (I'm asking because `git bisect` told me you introduced the changes in 2d21fcc9f683c1089522a1781318a045321fb617)

### DIFF
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -75,7 +75,7 @@
     // Sort helper that support null and undefined
     function ltHelper(prop1, prop2, equal) {
 
-      if (typeof(prop1) === "boolean" && typeof(prop2) === "boolean") {
+      if ((prop1 === true || prop1 === false) && (prop2 === true || prop2 === false)) {
         if (equal) {
           return prop1 === prop2;
         } else {
@@ -108,7 +108,7 @@
 
     function gtHelper(prop1, prop2, equal) {
 
-      if (typeof(prop1) === "boolean" && typeof(prop2) === "boolean") {
+      if ((prop1 === true || prop1 === false) && (prop2 === true || prop2 === false)) {
         if (equal) {
           return prop1 === prop2;
         } else {
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -95,6 +95,17 @@
         if (prop2 === undefined || prop2 === null || prop1 === true || prop2 === false) {
           return false;
         }
+
+        if (prop1 < prop2) {
+          return true;
+        }
+
+        if (prop1 > prop2) {
+          return false;
+        }
+
+        // not lt and and not gt so equality assumed-- this ordering of tests is date compatible
+        return equal;
       }
 
       if (prop1 < prop2) {
@@ -131,6 +142,17 @@
         if (prop2 === undefined || prop2 === null || prop1 === true || prop2 === false) {
           return true;
         }
+
+        if (prop1 > prop2) {
+          return true;
+        }
+
+        if (prop1 < prop2) {
+          return false;
+        }
+
+        // not lt and and not gt so equality assumed-- this ordering of tests is date compatible
+        return equal;
       }
 
       if (prop1 > prop2) {
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -75,23 +75,26 @@
     // Sort helper that support null and undefined
     function ltHelper(prop1, prop2, equal) {
 
-      if ((prop1 === true || prop1 === false) && (prop2 === true || prop2 === false)) {
-        if (equal) {
-          return prop1 === prop2;
-        } else {
-          if (prop1) {
-            return false;
+      // 'falsy' and Boolean handling
+      if (!prop1 || !prop2 || prop1 === true || prop2 === true) {
+        if ((prop1 === true || prop1 === false) && (prop2 === true || prop2 === false)) {
+          if (equal) {
+            return prop1 === prop2;
           } else {
-            return prop2;
+            if (prop1) {
+              return false;
+            } else {
+              return prop2;
+            }
           }
         }
-      }
 
-      if (prop1 === undefined || prop1 === null || prop1 === false || prop2 === true) {
-        return true;
-      }
-      if (prop2 === undefined || prop2 === null || prop1 === true || prop2 === false) {
-        return false;
+        if (prop1 === undefined || prop1 === null || prop1 === false || prop2 === true) {
+          return true;
+        }
+        if (prop2 === undefined || prop2 === null || prop1 === true || prop2 === false) {
+          return false;
+        }
       }
 
       if (prop1 < prop2) {
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -74,6 +74,19 @@
 
     // Sort helper that support null and undefined
     function ltHelper(prop1, prop2, equal) {
+
+      if (typeof(prop1) === "boolean" && typeof(prop2) === "boolean") {
+        if (equal) {
+          return prop1 === prop2;
+        } else {
+          if (prop1) {
+            return false;
+          } else {
+            return prop2;
+          }
+        }
+      }
+
       if (prop1 === undefined || prop1 === null || prop1 === false || prop2 === true) {
         return true;
       }
@@ -94,6 +107,19 @@
     }
 
     function gtHelper(prop1, prop2, equal) {
+
+      if (typeof(prop1) === "boolean" && typeof(prop2) === "boolean") {
+        if (equal) {
+          return prop1 === prop2;
+        } else {
+          if (prop1) {
+            return !prop2;
+          } else {
+            return false;
+          }
+        }
+      }
+
       if (prop1 === undefined || prop1 === null || prop1 === false || prop2 === true) {
         return false;
       }
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -108,23 +108,26 @@
 
     function gtHelper(prop1, prop2, equal) {
 
-      if ((prop1 === true || prop1 === false) && (prop2 === true || prop2 === false)) {
-        if (equal) {
-          return prop1 === prop2;
-        } else {
-          if (prop1) {
-            return !prop2;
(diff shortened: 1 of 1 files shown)


alexandernst:
Ah, I forgot to mention. This patch will honor the `equal` (optional) argument.

obeliskos:
This code path is very performance sensitive, so I would need to verify that overhead of typeof isn't a significant slowdown.  

I would not be surprised if your logic is spot on, but we should probably balance this against efficiency.  

I don't think there would be a problem with reversing the evaluations, as true > false makes sense.   Would that meet your truth table and be sufficient? 

I will try to find time to look this over this weekend.

alexandernst:
I'm not sure we can get past this without actually checking the type of the props. That codepath must check and properly evaluate boolean props, but make it in such a way that the logic table is applied only if both props are boolean.

Let me write a simple jsperf to check the penalty performance of typeof.

alexandernst:
Indeed, `typeof` is extremely expensive. After some tests I was nearly 50% performance penalty. Anyways, we can just strict-check for the type, that will make the code a little bit faster.

Check my tests: http://jsperf.com/typeof-perf/3

alexandernst:
We're completely overthinking this. The helper functions are a perfect candidate for memoization, which if I'm not wrong, will break all perf tests.

Basically, the first run with "new" props the comparator will be kind of slow, but next time it gets called with the same arguments, it will just return the result it returned the first time, without running all the code in the comparator.

What do you think about this? Should we give it a try?

alexandernst:
Indeed, as I thought, memoization just kicks everything else: http://jsperf.com/typeof-perf/3

obeliskos:
Hmm, interesting.   Might be interesting to experiment with memoization outside master branch just to guage effects.

My general feel is that we would run a query against... say a 5,000 document collection and let's say it's not indexed... so that may call ltHelper/gtHelper 5000 times for a single query, which will memorize 5000 strings/dates/integers globally?  

Without knowing if its a small database or one that is pushing memory limits of browser/nodejs its hard to say if memory is more valuable than cpu.  My general approach would be to make options which increase memory use 'opt-in'... and I don't know if you can do that here.

Feel free to experiment, I will look into it over the weekend.

alexandernst:
Not necessarily. If there are 5000 non-indexed documents and a query is ran, `ltHelper/gtHelper` will be called 5000 times, but the number of memoized values will depend on the props. So, le'ts say I have 5000 documents like this one:

    { name: 'Foo': has_car: true }
    { name: 'Bar': has_car: false }
    { name: 'ABC': has_car: true }
    { name: 'XYZ': has_car: false }

If a query to find only the people that don't have a car, all the possible combinations of `has_car` (that will be used to call `ltHelper/gtHelper`) are `true, false` and `false, true`, which means we'll end up with 4 

```
(true, false, true)
(true, false, false)
(false, true, false)
(false, true, true)
```

(including the third optional argument) memoized items for 5000 documents.

Let's make it a little bit more complex.

    { name: 'Foo': age: 34 }
    { name: 'Bar': age: 31 }
    { name: 'ABC': age: 33 }
    { name: 'XYZ': age: 35 }
    ...

If a find with sort-by-age query is issued, `ltHelper/gtHelper` will be called with the age of each person. There will be multiple combinations, but chances that 2 or more people with the same age exist in the database is very big, which will result in a speedup (because of the memoization).

Summing up, the worst case scenario where absolutely all values that are being compared are different, the number of memoized elements (including the third, optional argument) will be `binomial(N, 2) * 2 * 2` (being `N` the number of values).

That means that in a collection of 5000 elements where absolutely none of the compared values is equal to the others, we'd have [49.990.000](https://www.wolframalpha.com/input/?i=binomial(5000,+2)+*+2+*+2) memoized values in an object of objects.

Let's keep pulling numbers. The worst of the worst case scenario, the compared values are strings. And each string is, on average, 50 bytes long.

(the half of the total number is for the strings, the other half is for the boolean third argument)
`( (24995000 * 50) + (24995000 * 1) ) / 1024 / 1024 = 1215mb (1.2 gigabytes)`

That it still pretty much amazing, given the fact that the comparison won't take any time at all.

More numbers:

A digit (decimal or float) in JavaScript has the maximum size of 64 bits (implementation specific). This means that, at most, we'd have

`( (24995000 * 8) + (24995000 * 1) ) / 1024 / 1024 = 214mb (0.2 gigabytes)`

All that said, that is the worst case scenario, were none of the values equals to any of the other values.

The memoization could be limited to digits and booleans, which would still speed the process, and won't require so much memory. It could be also limited to certain number of memoized items and once reached that number, the memoization object won't be filled with more values.

How do you want me to proceed from here. What do you think could best fit Loki?

alexandernst:
Meanwhile I made another commit that makes the code a bit faster. This needs to get in as the code in it's current state just doesn't work fine. From here on, we need to discuss how to achieve the best possible performance.

obeliskos:
I had a chance to look over the code (both yours and ours) and think I might offer some input to make sure this is on right path.  So these helper functions delve into some ugly places and Booleans get dragged into it a little.  A lot of the ugliness is that (being a document database) we can't guarantee properties being compared are of same type or even exist.

So i'm thinking we can category limit some of the ugliness by testing for ('falsy' || strict true) in either prop1 or prop2 and then handling those conditions, suboptimizing as necessary.

So for starters can you wrap your (optimized) Boolean tests, along with the (existing) two 'if' blocks that deal with undefined and null (and bool) in a larger 'if' block structured like : 
```javascript
// 'falsy' and Boolean handling
if (!prop1 || !prop2 || prop1 === true || prop2 === true) {
...
}
```

I believe that may minimize evaluations necessary for all other main path evaluations and provide a more logical subcomposition for further optimizations.

Just trying to keep some semblance of a 'big picture' to these routines as they have evolved from many specific changes like this and they are somewhat losing coherence.

alexandernst:
You mean like this? @obeliskos

obeliskos:
Yea, after thinking about it we might need to 'duplicate' the final three tests within that block as well since  0 (integer) and "" (empty strings) get pulled in as well.

And do the same for the ltHelper?

alexandernst:
Done the `ltHelper`.

What do you mean with "the final three tests"?

obeliskos:
Actually 2 tests and the default return : 
```javascript
  if (prop1 < prop2) { 
    return true; 
  } 
 
  if (prop1 > prop2) { 
    return false; 
  } 
 
  // not lt and and not gt so equality assumed-- this ordering of tests is date compatible 
  return equal; 
```
duplicate that code within the block as well as leaving outside the block so that empty strings and 0 integers can also fallback to that logic.

alexandernst:
@obeliskos I think we're done now :)
Anyways, following this I think I'll write some specs for `lt/gt Helpers`. This is a very sensitive code path.

obeliskos:
Yes please do, as you can imagine the combinations are many and sometimes ambiguous.  Half the effort is figuring out how to figure out how to compartmentalize the tests for best performance.

Thanks for your patience

alexandernst:
Sure, no problem.
About the memoization, do you know what we should do in that area?

obeliskos:
As for applying it to ltHelper and gtHelper we would need to do it as some sort of op-in experimental mode.  

My guess is that the following issues should be addressed for such a feature (at least I would address them in my own experiments) : 
- benchmarks : our current benchmarks (benchmark/benchmark.js) apply mostly $eq tests so you might need to create a benchmark to gauge effectiveness accurately.
- IIFE global : the ltHelper and gtHelper routines are global in scope to the iife, so to add a loki or collection 'option' for turning this on or off we could need to store copy globally or pass down the call stack.  While experimenting I would rather just keep global experimentalMode=false which we can switch on by editing lokijs (and can hopefully be optimized out or in by jitter).
- Estimate or instrument memory overhead : we (or you) should be able at least while experimenting to have actual running estimate of memory overhead.  maybe we could use node.js process.memoryUsage() in benchmarks to dump this before and after a 'memoize' bench test.
- code staging : is this best done on master branch or should we move this to 2.0 experimental branch or techfort's 'skrymir' repo or personal fork until all the pieces and impacts can be gauged.

Any thoughts or concerns on these?

alexandernst:
No concerns right now, but I'm sure I'll have some when I'm knee-deep in that refactoring :)
Ok, I'll start in a branch in my fork, see how things go, then make some kind of a PR here so we can all discuss it.

obeliskos:
Sounds great... sure thing

PR #365, opened 2016-02-29
Fix corner case for glHelper

alexandernst:
Make glHelper honor `equal` argument.


alexandernst:
I think this is done @obeliskos @techfort 
Ready for a review!

NachE:
Hi! How about the status of this pull? I find this code useful for my project

obeliskos:
Hey, sorry it has taken me so long to get around to this... time has been limited.

I see what you are doing there, and I never saw !! type coercion before, but seeing as though this is internal helper only, it would probably be best to replace all calling functions which omit this and add 'false' param to correctly initialize parameters.

I realize you probably wanted to keep scope simpler but if we can eek out a millisecond here or there by not doing type coercion then I would rather just make sure the method is properly invoked.

Either you do that, or I can do it and I assume your new specs will start passing as well?

alexandernst:
@obeliskos No problem!

Please do, I'm short of time today. Indeed, making sure everything is `false` instead of doing `!!`will make things a little bit faster.

And yes, specs will pass once this is merged.

alexandernst:
@obeliskos A had some spare time today, so I updated the code as you requested :+1:

obeliskos:
Ah, i did that yesterday after you saying for me to go ahead and do that.

Was there anything else in the p.r. aside from 'equal' param handling?  I figured the #366  would be where maybe you needed to update something since it still does not pass.

Let me know where these stand at the moment... Thanks!

obeliskos:
If there are other minor changes needed to these helpers, can you add those changes to #366 instead? Confusing dealing with these separately... thanks.

alexandernst:
Ah, sorry, I thought you hadn't done it yet.

The equal param was the only missing thing. Tests dont pass yet since this
PR isn't merged. Once this PR is merged, I'll trigger a restart in Travis
in #366.

obeliskos:
Please make your equal param change within #366 itself.  Feels like these two (dependent) pr's should have been a single one to begin with.   With that done and passing I would pass it and just close this one.

alexandernst:
@obeliskos Just the equal param change won't make #366 pass. This entire PR is required for #366 to pass. Do you want me to cherry-pick all the commits that are here in #366?

obeliskos:
It would appear you would need to maybe cherry pick two lines of code?  Am I missing anything? Its split out into so many sub-commits its hard to get a grand 'diff compared to master' report.

obeliskos:
Ok looks like you can append ".diff" to the url of any github PR url to get a comprehensive diff of the (whole) PR compared to (master?).   Seems it still is comparing old version of master though since many of the changes have already been changed.

Still its clean enough I could commit this if you prefer.

alexandernst:
@obeliskos To see a full diff of this branch compared to master, just use https://github.com/techfort/LokiJS/pull/365/files

Yes, committing this should make green #366

obeliskos:
That view is even better, thanks.  

Yea I'm guessing maybe you haven't merged recent commit into this pr so its showing some redundant fixes which are already in master but it should merge well enough.

PR #366, opened 2016-02-29
Create specs for $lt and $gt (among others) comparators

alexandernst:


alexandernst:
@obeliskos Please note that in order to test the comparators without making funny things with the code, I exposed `LokiOps`. It shouldn't matter, but I thought I'd mention it just in case.

Also note that the specs won't pass because `gt` and `lt` comparators are still buggy. Tests should be fixed after merging https://github.com/techfort/LokiJS/pull/365

alexandernst:
@obeliskos Some "crazy" tests are missing. "Crazy" as in "comparing anything with anything".
I'm not sure what should those return. For example `true > 'foobar'`? `new Date(2015) > 42`?
Should those just return whatever the JS interpreter returns?

techfort:
@alexandernst @obeliskos if i can interject, if you compare a boolean to a string i'd throw a `new Error('change jobs, dude.')` . Having said that, we can assume people know what they are doing so the result should be - as @alexandernst said - what the interpreter returns (wilth all the caveats of JS, such as `{} + []` being `0`, but `[] + []` being `''` and so on...)

alexandernst:
I'm almost sure the current code (not this specs, but the comparators) covers all possible cases of "crazy" combinations. And when I say "almost" I actually mean "I haven't found a combination that doesn't return what it should".

Are we good with this PR? Remember that this mist be merge only after https://github.com/techfort/LokiJS/pull/365

alexandernst:
Ok, now that #365 got merged, I'll restart travis here

obeliskos:
Whew, thanks for these tests (and fixes).

alexandernst:
No problem :)

PR #374, opened 2016-03-12
fix Collection.get method

KostiantynKopytov:
wrong comparison between object and number

### DIFF
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -3925,7 +3925,7 @@
       while (data[min] < data[max]) {
         mid = (min + max) >> 1;
 
-        if (data[mid] < id) {
+        if (data[mid].$loki < id) {
           min = mid + 1;
         } else {
           max = mid;


obeliskos:
'data' (in that line of code) is actually a local variable pointing to idIndex, which is an array of integers.

If you go to : [Loki Sandbox page](https://rawgit.com/techfort/LokiJS/master/examples/sandbox/LokiSandbox.htm) , you can click on 'Object Inspector' to drill down into all the properties (like) idIndex to see what kinds of data they store.

KostiantynKopytov:
Oh I see. Sorry for wrong pull request.

But still don't understand why not use idIndex as a dictionary instead to make this possible idIndex[someLokiIndex] -> get some data by id.

obeliskos:
No problem.   Refactoring of idIndex will be left to a 2.0 version change since this would break serialized file compatability.  

Performance is better with collection.by() than collection.get() but not by so large a margin that it can't wait for a 2.0 version (or at least minor version with autoupgrade logic).

PR #383, opened 2016-03-18
Fixed 'cloneMethod' option casing and clarified anonym() documentation

TeamworkGuy2:
- The Collection constructor references options.clonemethod (lower case) and options.cloneMethod (camel case), I'm assuming both references should be options.cloneMethod
- The anonym() method documentation indicates that you can pass in an indexes array, but the indexesArray is being passed as the Collection constructor 'options' parameter, so I changed the documentation to match, rather than changing the behavior


VladimirTechMan:
Good job! Thanks, @TeamworkGuy2.

PR #395, opened 2016-04-04
Node v5.10.0 update, specified digest

sertaconay:
crypto.pbkdf2 without specifying a digest is deprecated. (Node v5.10.0)
If digest is not specified, default of sha1 is used on earlier Node versions. So, i didn't want to break anyone's app and set digest as sha1.

The error it gave on v5.10.0 was:
`DeprecationWarning: crypto.pbkdf2 without specifying a digest is deprecated`


obeliskos:
Excellent, thanks!

PR #406, opened 2016-04-24
Exposed persistence adapters to Loki.persistenceAdapters.

### LINKED ISSUE #404: Expose the built in persistenceAdapters.
Hey there!

As the title suggests I want to discuss if you could expose the Loki built in persistence adapters to something like:
```javascript
loki.adapters //  => {fs: /* constructor function of fs adapter */, localStorage: ...}
/* or */
loki.persistenceAdapters // => like above
```
To make this more clear: the ``loki`` variable in the example above is the constructor not an instance of lokijs.

My use case is a "proxy" adapter that serializes the database to [messagepack](http://msgpack.org/) instead of json. And I would like to use the built in adapters instead if writing an own one for fs and local storage.

Greetings from Germany

Kanaye:
Closes #404.

### DIFF
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -689,7 +689,7 @@
      * @param {boolean} options.autoload - enables autoload on loki instantiation
      * @param {function} options.autoloadCallback - user callback called after database load
      * @param {adapter} options.adapter - an instance of a loki persistence adapter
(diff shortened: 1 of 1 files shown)


techfort:
cheers @Kanaye

PR #414, opened 2016-05-07
Added automatic support for nativescript framework

stefalda:
I created a new adapter to support loading and saving of the database in a nativescript application.
Because it's possible to detect if the runtime environment is nativescript I've slightly modified the Lokijs file to automatically select the nativescript adapter if no adapter is passed in the options.

Performance seems quite good both in the iOS and in the Android app.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 
 LokiJS is a document oriented database written in javascript, published under MIT License.
 Its purpose is to store javascript objects as documents in a nosql fashion and retrieve them with a similar mechanism.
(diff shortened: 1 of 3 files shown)


techfort:
@stefalda ciao! my guess is that the build fails because LokiJS is currently written in ES5 and you used the `const` keyword. I might have missed something else though, the travis log doesn't seem very meaningful in this case. Can you re-check?

stefalda:
Ciao Joe! Now it seems fixed :-) 
Thank you

PR #441, opened 2016-06-27
add: callback for database delete

### LINKED ISSUE #440: IndexedDB Adapter - delete callback?
Hi,

Correct me if i'm wrong. I can't seem to [find the callback](https://github.com/techfort/LokiJS/blob/master/src/loki-indexed-adapter.js#L180-L204) mechanism for deleting a database from IndexedDB?

There is a comment there written, which I couldn't comprehend

```js
// lazy open/create db reference so dont -need- callback in constructor
```

Any such reason why it isn't there?

vasumahesh1:
Fixes #440

### DIFF
--- a/src/loki-indexed-adapter.js
+++ b/src/loki-indexed-adapter.js
@@ -172,22 +172,26 @@
      *
      * @example
      * // DELETE DATABASE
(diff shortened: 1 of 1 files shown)


techfort:
cheers @vasumahesh1 !

obeliskos:
excellent, thanks!

PR #442, opened 2016-07-02
Recursively deep dive using dot notation

markgillnz:
Allows finding documents with properties that are multiple levels deep within objects and arrays, using notation such as 'grandparent.parent.child'


techfort:
@markgillnz awesome stuff. Didn't have time to see if this impacts performance much but looking at the commit it should not. Thanks!

Issue #439, opened 2016-06-20
Bug in DynamicView...


deenfoxx:
I found a bug (or misunderstanding) in DynamicView which I was easily able to duplicate on the Loki Sandbox using most of the demo code...

In a nutshell... If a DynamicView does NOT have a filter, it appears to force one into existence as soon as an item is updated. Try copy-pasting into https://rawgit.com/techfort/LokiJS/master/examples/sandbox/LokiSandbox.htm

If I am just misunderstanding, perhaps you can clarify, because I thought I would be able to removeFilters() and be able to access the entire database as if I were accessing it directly. However, if I modify a single document through the primary db, suddenly I no longer can view anything except what was just updated.

```
var db = new loki('sandbox.db');

// Add a collection to the database
var items = db.addCollection('items');

// Add some documents to the collection
items.insert({ name : 'mjolnir', owner: 'thor', maker: 'dwarves' });
items.insert({ name : 'gungnir', owner: 'odin', maker: 'elves' });
items.insert({ name : 'tyrfing', owner: 'Svafrlami', maker: 'dwarves' });
items.insert({ name : 'draupnir', owner: 'odin', maker: 'elves' });

// Create a DynamicView with NO filter.
var dv = items.addDynamicView('query');
logText('GOOD... Shows everything, as expected with no filter.');
logObject(dv.data());

// Find and update an existing document
var tyrfing = items.findOne({'name': 'tyrfing'});
tyrfing.owner = 'arngrim';
items.update(tyrfing);

// Try and get Dynamic View again... Whoops?!
logText('BAD... No Filter applied, but only shows updated item. Lost all the others items.');
logObject(dv.data());

// Try to do the applyFind a little late?
dv.applyFind({ 'owner': 'odin' });
logText('This SHOULD be able to find odin items... nope...');
logObject(dv.data());

// This statement sends to Inspector
inspectObject(db);
```

deenfoxx:
Unrelated... but wanted to show off... Your LokiJS is working awesomely! Fast as hell even with some bad programming choices because I was rushing things yet still wanted it feature-rich. haha...

https://bestticketprinting.com/search-test/bachelor-party

Play with the search keywords, categories and filters... Uses AJAX to pull in search results live as you change keywords/categories/filters... FUN!

[...]

PR #443, opened 2016-07-05
DynamicView: Send notifications and update persistent data even if 'filterInitialized' is false on resultset

VladimirTechMan:
Unfortunately, the fix for #439 broke things a little bit (causing a regression in a recently released public version 1.4.0): When the `filterInitialized` flag on `resultset` remains `false`, it is still required to send the "rebuild" notifications to the subscribed listeners on the instances of `DynamicView` and update persistent data, if the related option is set on those instances. After fixing that part, we should be fine, I think.

@obeliskos, please, take a quick look when you have time. Based on my internal tests, we should be good after applying the changes that I propose here. (Though the way that I am updating `resultdata` may not be the fastest yet – it may need more fine-tuning; but, at least, it will be correct now.)

BTW, looking at all the duplicate code fragments that either queue the sort phase or just send the 'rebuild' notifications when no sorting is required, I feel more and more like refactoring the `evaluateDocument` and `removeDocument` methods. As soon as I have a little more time in the coming weeks, I will take a closer look at that.

Looking at that code today, I have also realized that when the option for data persistence is enabled on an instance of `DynamicView`, we do not check and respect the cloning option at all: We always push direct references to the collection documents into `resultdata`.

### DIFF
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -3083,6 +3083,15 @@
     DynamicView.prototype.evaluateDocument = function (objIndex, isNew) {
       // if no filter applied yet, the result 'set' should remain 'everything'
       if (!this.resultset.filterInitialized) {
+        if (this.options.persistent) {
+          this.resultdata = this.resultset.data();
+        }
+        // need to re-sort to sort new document
+        if (this.sortFunction || this.sortCriteria) {
+          this.queueSortPhase();
+        } else {
+          this.queueRebuildEvent();
+        }
         return;
       }
 
@@ -3178,6 +3187,15 @@
     DynamicView.prototype.removeDocument = function (objIndex) {
       // if no filter applied yet, the result 'set' should remain 'everything'
       if (!this.resultset.filterInitialized) {
+        if (this.options.persistent) {
+          this.resultdata = this.resultset.data();
+        }
+        // in case changes to data altered a sort column
+        if (this.sortFunction || this.sortCriteria) {
+          this.queueSortPhase();
+        } else {
+          this.queueRebuildEvent();
+        }
         return;
       }
 


techfort:
@VladimirTechMan i merged this in the meantime because i'd rather know that there is an issue with performance than correctness. I will be announcing 1.4 today so this is definitely better, thanks!

obeliskos:
Hmm so do they need rebuild events when they haven't added a filter? I would be inclined to not build persistent until user calls data the first time.  I suppose we might say if (!initialized && !persistent && !sortCriteria) then return

Yea if we are going to provide a 'persistent' array which users can access with minimal rebuild we should do cloning into resultdata and do it 'up-front' or at least on first data call and evaluate methods.

I am having issues with my known email address (for a while now) through my web service provider so I will try to resolve that to allow better communication on this subject.  Until then we can communicate here.  I'll look this over more later tonight or so.

VladimirTechMan:
Thanks, @techfort, for the quick action on that. Yes, 1.4.1 is now available and I have updated it through npm this morning, thus I can keep using the general public version of LokiJS in my project for now. (@techfort, there are some other minor things that I wanted to address, including one related to another change made very recently, but those are less urgent for now and can wait.)

@obeliskos, thanks for letting me know about the temporary issues with the e-mail address that you shared previously. Using GitHub for communicating on this topic is fine with me. Now, to the topic itself:

I believe, the major thing for us here is to agree on the scope of the functionality that `DynamicView` covers (and is intended to cover). I understand that you likely only intended it to be a "configure once before the use" tool, initially. Then, last year we implemented the mechanism of named filters. (Partially, to address my needs in the specific missing functionality of Loki at that point of time – actually, all of those needs are still alive and that functionality is in use these days, on my side). 

Now, the named filters in `DynamicView` instances make it very natural and efficient to use dynamic views in a way where some filters can be applied and revoked "on the fly", as necessary. One natural "edge" use case of that being that the application may want to start with an "empty" (that is, without any filters in the chain) dynamic view, then apply (and possibly, later update or remove) some filters to it. An application doing that may still need to be notified whenever the contents of the view changes, upon adding, updating, or deleting some documents to / in the associated collection. Even if that reflects the whole "unfiltered" collection. `DynamicView` just acts as a universal view mechanism here, being truly dynamic both in terms of the documents "viewed" through it and in terms of the filter chains being changed or removed as necessary. (At least, that was my perception of dynamic views – was it going too far?)

There is no question that a developer can also handle the situation with an "occasionally empty" filter chain in a dynamic view by directly accessing the collection itself and listening to its events, associated with changing the documents contained. But that would force developers who need that functionality to implement yet another layer of abstraction, now on top of `DynamicView` and `Collection`. Where, to me, `DynamicView` can very naturally and efficiently cover those usages itself, without any drawbacks. @obeliskos, any objections to that, from your perspective?

Now, I am actually not concerned about the case when there is no sorting and persistence options at all (and no filters initialized yet). With the fix that I have implemented, that scenario will simply cause the "rebuild" events to fire periodically, as intended, depending on the configured minimal event interval. And the rest of required functionality in the `data` and `count` methods of `DynamicView` is just covered by the internal logic of `Resultset`, which already properly handles the situations where `filterInitialized` evaluates to `false`. No side effects of that, that I am ware of. The same story, indeed, in the case of a sorting option applied.

It is just that specific option for persisting the view's result data that "dictates" that `DynamicView` prepares its data as soon as it can, trying to guarantee a near-constant access time to them later, as opposed to the case of "calculate upon request" approach that we use by default. I currently only utilize that later approach (and I am quite happy with it so far, for my purposes). But I trust that some people do use the persistence option – and then, they might want to combine it with the "rebuild" event notifications in the way that I described above.

Let me know what you think.

obeliskos:
Ah yes thanks for reminding me how I would probably expect that to work without being preoccupied with internals.  

So my main issue was tracking a bug (#439) down and finding evaluateDocument() adding document (positions) to filteredrows on an Resultset with filterInitialized set to false.  To avoid evaluate adding those docs I just short circuited.  As long as we detect and cover preserving the 'resultset' state to a meaningful state we can elevate higher state management at the dynamic view level.  I do like the simple passthrough if non-persistent though... in which case cloning and sorting are done late, on call to data(). 

As you mention, events probably should be emitted for all variations, even the non-persistent scenario I mentioned above.

We probably should have more unit tests for these scenarios.  That would probably prevent either of our expectations from being transparent to the other.  I have set up spec/generic/dynamicview.js which we can expand out a few higher level state tests.

So my main concern is to keep the state management clean and as simple as possible.  The resultset state (as mentioned) is only really concerned with whether it is initialized (to determine whether to run queries against data[] or filteredrows), so whatever dynamic view state management we set up, it needs to make sure resultset doesn't get confused if we ask it to give us data. 

Email is also working again, should you wish to pursue more granular discussions to work out any details.

VladimirTechMan:
@obeliskos, sorry for a delayed response on that from my side. Yeah, we can shift part of this discussion back to our e-mails if required. I fully agree on adding more unit tests for that (and some other parts of Loki's functionality) – we'll need to do that going forward.

As for the correctness of the state management of `Resultset` that we are doing inside `DynamicView`. The good news, as I mentioned it in my reply above, is that resultset correctly checks and handles the situation with `filterInitialized` evaluating to `false` inside its methods. That helps us here – and like you noted, gladly, that allows us to do a simple pass-through, safely, when the persistence option is not involved. Benefiting from later "lazy" management of data and sorting.

From my perspective, the two things that (may) start to make it more challenging for us to reason about whether that part of LokiJS would be functioning efficiently and correctly with different code paths involved – those two things are:
- The replicated fragments of code responsible for handling sorting and notifications and persistence, when enabled, inside `evaluateDocument` and `remoteDocument`. Those are getting spread out through different code branches, more and more. Not a show stopper, can lead to regressions like the one I reported (the proper suit of test units will help us, but we may also think about some refactoring to that code logic – too much of a repetition, to my developer's eye).
- A little bit more "fundamental" aspect here is, that we actually manipulate the state of the nested resultset (`filterInitialized` and other state properties of it) externally from the code inside `DynamicView`. That shifts the responsibility for maintaining the whole state and behavior of resultset to dynamic views, on one hand. And, on the hand, it equally makes us guessing and double-checking if the code of resultset would be ready to "recover" properly if its state has been (suddenly) made inconsistent by the client dynamic view. Both parts point to a need for a potential update to the design of that interfacing part between `Resultset` and `DynamicView`, eventually.

Okay, the rest we can cover in the e-mails, I guess.

obeliskos:
I agree on reusing duplicate code blocks... and our unit tests should not necessarily need to test all duplicate code paths, just outcome.

Yes, before I forget it the actual issue I ran into can be seen here : 
https://github.com/techfort/LokiJS/blob/e2e00996345969b661b63c57a0302099e5a4f439/src/lokijs.js#L2342-L2343

The next line sees that filteredrows actually contains values and uses that.  Our evaluateDocument thought it needed to add to filteredRows even though we were in a (no filter/all records) state.

Ok, yes we can communicate by email or possibly try to coordinate a time for meeting up on the loki Gitter page which I have yet to use.

PR #470, opened 2016-08-28
fix for circular structure error when setting ttl - fixes #469

### LINKED ISSUE #469: TypeError: Converting circular structure to JSON
When setting `ttl` and `ttlInterval` on a new collection, I am getting this error:
`TypeError: Converting circular structure to JSON`

It seems that the `ttl` key should be added to `Loki.prototype.serializeReplacer`. When I add that, the error goes away:

```
    Loki.prototype.serializeReplacer = function (key, value) {
      switch (key) {
      case 'autosaveHandle':
      case 'persistenceAdapter':
      case 'constraints':
      case 'ttl':
        return null;
      default:
        return value;
      }
    };
```

dbubs:
When `ttl` is enabled on a new collection, you will get a `TypeError: Converting circular structure to JSON ` error. This PR will fix it.

### DIFF
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -906,6 +906,7 @@
       case 'autosaveHandle':
       case 'persistenceAdapter':
       case 'constraints':
+      case 'ttl':
         return null;
       default:
         return value;


techfort:
@dbubs cheers!!

Issue #465, opened 2016-08-24
How to observe changes by query?


iovglaz:
In Meteor minimongo, we can do something like:

``` js
var cursor = collection.find({ $gt: { abc: 120 } });
cursor.observe({ updated: function(oldDoc, newDoc) {});
cursor.insert({ abc: 123 });
```

In LokiJS i see `collection.on`, but how can we keep track of such an event for a particular data?

Or how do something like this?:

``` js
var doc = collection.insert({ abc: 123 });
collection.on('remove', function(data) {
  if (collection.match({ $gt: { abc: 120 } }, data)) {
    console.log(true);
  }
});
collection.remove(doc);
```

obeliskos:
It would seem [here](https://github.com/techfort/LokiJS/blob/e9a2e51d104a0c2a420e42182d82b5ea55d17a39/src/lokijs.js#L4151) that our emitted event is called 'delete'... maybe in future we can rename this event to remove.

Since it looks like you are not attempting to affect outcome of remove, just monitor it... your logic looks good.   We don't have a collection.match method but if i understand logic correctly, you don't need it?  Maybe something like : 

```javascript
var doc = collection.insert({ abc: 123 });
collection.on('delete', function(data) {
  // in case your doc is missing property, don't throw js error
  if (data.hasOwnProperty('abc') && data.abc > 120) {
    console.log(true);
  }
});
collection.remove(doc);
```

iovglaz:
I think this is a joke... You are offer to write an universal interpreter of any queries?

In lokijs I not found two things for the work triggers.

##### Old state of the document on `update` event.
I do not know how to solve it ...

##### Checks the document for compliance with the request manually.
Of course we can write a crutch.
Create a temporary collection with out data and then make a find query for them.
But... It awful.

-----

There's something that could be done almost without any loss of performance.

We can assign for all operations of a numeric value by the counter. This may be optional feature. If this number is transmitted to the event, we can be accurately associate the request with the event on it. We do not have to check the document for compliance with the request, and so on.

obeliskos:
I'm not sure I understand what your asking or what your points are...   Perhaps you can rephrase the important points.

Regarding 'old state of document on update event' : while our current implementation only passes a single document, this could be modified to pass original as well as new document as long as you have the collection 'clone' option passed as true in [constructor](https://rawgit.com/techfort/LokiJS/master/jsdoc/Loki.html#addCollection).

[...]

PR #471, opened 2016-08-29
Added support for old document state on update event if options.clone is true

ivansglazunov:
Associated with https://github.com/techfort/LokiJS/issues/465.


ivansglazunov:
This change adds a second argument containing a clone of the original document, if `options.clone` is `true`.
```js
var collection = db.addCollection('name', { clone: true });
collection.on('update', function(updated, old) {});
```

obeliskos:
Thanks, i must have missed that code path.  I'll add a unit test for it to ensure it keeps working.

obeliskos:
Ok i see you did this on 2.0.0 branch... 

I think this p.r. clued me in as to what you and/or #465 were requesting, but think I see a few issues with the implementation.

First i will assume two net goals : 
- the object/document instance you pass to update() will never be used internally in a clone situation.
- the 'on update' event only cares about a facsimile (copy) of the original (internal) and after (what was passed to update)

We really don't want the user having references to documents that we are using internally (if cloning is enabled).   I see now that we weren't doing that clone for updates and I or you need to add this to main branch.   I believe you are only cloning the emitted object, we need to clone the object before it gets stored in the collection... we can then either make a different clone for the emit or just emit the user (passed in) ref... probably more optimal to pass user ref for now (unless we can easily check if anyone is listening before doing performance expensive 2nd clone).

I think I will go ahead and fix the first of the two goals (on master branch) because I thought that was already working... and i'll add unit test.  At that point, if you would like you can submit a p.r. on master to add the extra functionality to the emit.

ivansglazunov:
https://github.com/techfort/LokiJS/issues/465
This issue has been another problem. Comparison of the event with selector.
I do not know how best to implement it here. II see one solution, but they are not perfect.
Independently verify the document on conformity to selector.
```js
collection.on('insert', function(document) {
  verify(document, { $or: [ {abc: 123}, {abc: 456} ] }); // Boolean
});
```
Then it will be possible to listen to events, not the entire database, but only partially.
It will load but will eliminate the need to change the database logic.
Is there a recommendation where to get the ingredients for such a function?

obeliskos:
So if the evaluation occurs within your callback this is just expressing your filter as an object rather than javascript property evaluations.  Your event wiMaybe something we can throw off LokiOps.

I am guessing this is a feature already implemented somewhere else?  If so perhaps a link so I understand the functionality implied by this.

So if verify returns Boolean (and I see no other user callback) am I to assume that you would use this such as : 

```javascript
collection.on('insert', function(document) {
  var docMatch = verify(document, { $or: [ {abc: 123}, {abc: 456} ] });
  if (docMatch) {
    // perform some actions on documents which are limited in scope to our 'verify'
    console.dir(docMatch);
  }
});

**Comment by ivansglazunov** [comment]
Yes. Please look at the class [Meteor Minimongo Cursor](http://docs.meteor.com/api/collections.html#mongo_cursor). It is very convenient to use when working with client collections, for example in React. To make such a class can be almost not dependent on the LokiJS, easily adaptable to different databases. To implement this functionality based on LokiJS only need to subscribe to changes to the data by LokiJS query selector. This feature can be implemented completely separately, re-interpreting the query language is similar to Mongo, etc. Or we can use something embedded in LokiJS if this will be available as a method. Which path is more likely?

**Comment by obeliskos** [comment]
So if we already have something like [this](https://gist.githubusercontent.com/obeliskos/7e12d6e9fe9d384324b78d549c5adfd5/raw/8319ca806204f09607e6917830c6875d94463ecd/match1), maybe we can expand other ops to be more comprehensive document level evaluations.

**Comment by ivansglazunov** [comment]
What's better:
###### Embed cursors functional on LokiJS level.
```js
// method name in question
var cursor = collection.getCursor(selector);
cursor.on('inserted', function(newDoc) {});
cursor.on('updated', function(newDoc, oldDoc) {});
cursor.on('removed', function(oldDoc) {});
```
###### Implement cursors based on the adapters and not dependent on the database.
```js
import { Cursor } from 'some-cursor-class-package';
import { Cursor as LokiJSCursor } from 'adapted-lokijs-cursor';
LokiJSCursor instanceof Cursor; // true
var cursor = new LokiJSCursor(collection, selector);
cursor.on('inserted', function(newDoc) {});
cursor.on('updated', function(newDoc, oldDoc) {});
cursor.on('removed', function(oldDoc) {});
// and then integrate it into some LokiJS method as getCursor
```

Maybe it's a matter of policy the use of external packages?

-----

PS. And... why LokiJS not use js `class` as in [airbnb](https://github.com/airbnb/javascript/blob/master/README.md#classes--constructors)? There is a reason not to use them?

ivansglazunov:
Independent cursors allow a class to build application logic more independently from the database. It will be possible to make requests to the mongo, lokijs, sql differently and work with the results is the same everywhere.

obeliskos:
well for 1.x.x (master) branch we prefer minimal to no dependencies and es5 compact with no transpile. 

2.x.x branch is open for as much refactor as needed.  Need to make sure its up to date with recent changes though.  @techfort, @VladimirTechMan , and @alexandernst might also want to offer opinions on 2.x.x refactorings (if they have preference and/or time ATM).

I'll need to read up on the cursors ... sounds similar to dynamic view provided we added new events..

ivansglazunov:
Yes, it's almost one and the same. The only difference is that the cursor in the meteor does not know the logic of the database. Not able to sort, limit and skip. If this functionality will be in dinamic view, it would be optimal.

Not sure whether this "unified logic of cursor" independent from databases. The question is not quite about LokiJS, but still, what do you think about it? Is this worth doing?

obeliskos:
So independent unified cursor implies to me you are willing to give up implementation specific optimizations and/or you will have to provide an interface for other databases to 'adapt' to?  Would you need adapter between unified cursor and loki or would one know about the other?

Would query engine be contained within cursor component?

Would notifications of changes require observability or interface/adapter?  cursor.on('insert') would have to be notified itself unless it could observe or poll some datasource.

ivansglazunov:
> Would query engine be contained within cursor component?

Query will be transferred to the performance database.

Class cursor can be simultaneously play the role of interface. His methods can be documented but not filled. It will need to inherit so that he would have played the role of the cursor and the adapter between the cursor and the database.
In the end it not (cursor(adapter(database))) but (cursor=adapter(database)).

Plus tests inherited adaptive cursor to specific input and output data as well as the fact the correct response to the changes.

Issue #474, opened 2016-09-08
Factory function based inflation


Cortys:
### Basic proposal

Currently Loki only supports constructor based object inflation. I think adding an option for a factory function based inflation would make sense.

So that in addition to:

``` js
store.loadDatabase({
    collectionName: {
        proto: SomeClass,
        inflate: (src, dst) => Object.assign(src, dst)
    }
});
```

something like this would also be possible:

``` js
store.loadDatabase({
    collectionName: {
        inflate: src => new SomeClass(dst)
    }
});
```

This would enable the use of inflation with a non class based data model.

If there is any interest, I can provide a pull request for this.
### Optional extension

Should this idea be realized, I think it's also worth considering an async extension of the feature.
This would make it possible to return a Promise to an inflated collection entry from the `inflate` function:

``` js
store.loadDatabase({
    collectionName: {
        inflate: src => {
            return loadSomethingAsync(src.url).then(dst => Object.assign(dst, src));
        }
    }
}).then(() => console.log("All objects successfully inflated."));
```

Since this would introduce Promises to Loki this would probably require API changes at various different places to keep everything consistent (basically replacing all callbacks with Promises), so it might not be feasible to do this until 2.0.

I'd love to hear your opinions.

obeliskos:
So inflation would occur on load and would populate our internal arrays using a factory (function?)  provided at the time of db construction?  If so, when the database saves with those factory created objects and then re-inflates later it is expected to produce the same type of object with similar properties?  Sorry ;) Just trying to make sure I understand correctly. 

I think I see the value of having a wide variety of methods for inflation, either for internal object instances or for object instances returned to the user from queries.  

Promises would seem to be one of the next refactorings we could use to adapt to more evergreen environments.  Performance will still be priority, but this should be fairly easily measurable.  

I am open to community consensus as to logistics and timeframes for opening up to newer api's and toolchains.  Any ideas on whether an effective 'beta'/development branch would need its own npm package or beta versioning?   Not sure if that's needed but open to opinions/suggestions.  Until such time as a consensus or migration strategy, updating of test cases and documentation for production users it should likely be on 2.0.0.

Cortys:
Yes exactly. That's what I meant.

I think the basic proposal without async support via Promises could easily be added without breaking anything, since API-wise it just allows users to let the `proto` property be undefined (which was forbidden before) and handle object instantiation themselves.

obeliskos:
[...]

PR #475, opened 2016-09-11
Support collection inflation without constructor

Cortys:
As discussed in #474.

Implementation includes spec.


obeliskos:
Awesome... and thanks for the unit test!

PR #490, opened 2016-10-19
refactor(insert): emit once for bulk preinserts and inserts

Thrilleratplay:
When inserting an array of objects, emit `insert` and `pre-insert` once instead for each object.


techfort:
@Thrilleratplay thanks, good stuff!

PR #493, opened 2016-10-20
add/remove array of event names

Thrilleratplay:
Allow `LokiEventEmitter.prototype.on` and `LokiEventEmitter.prototype.removeListener` to accept arrays of event names to be applied to a given listener.

Also, trim some trailing white space.

### DIFF
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -514,13 +514,21 @@
 
     /**
      * on(eventName, listener) - adds a listener to the queue of callbacks associated to an event
-     * @param {string} eventName - the name of the event to listen to
+     * @param {string|string[]} eventName - the name(s) of the event(s) to listen to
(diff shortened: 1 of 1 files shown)


obeliskos:
Hey, thanks... wouldn't you need to do something like this : (?)

```javascript
var event, 
  self=this; 
 
		self.on(currentEventName, listen); 
	}); 
} 
 ...
```

That should be scope visible way to use loki instance 'this' within forEach.  I think the way it is now is calling a (non-instance) prototype.   I guess alternatives are use 'call' along with 'self' on prototype or change forEach to for and just use 'this'.  

If I'm missing something let me know and I will test later today for my own benefit... otherwise I think both 'on' and 'removeListener' may need that minor rework.

Thrilleratplay:
@obeliskos  You are absolutely right.  I made the update.

PR #497, opened 2016-10-22
Migrated to Promise based API

### LINKED ISSUE #482: Switch to Promise based async methods
Currently all async methods of the API use callbacks. I want to propose moving to a Promise based implementation.

**To clarify:**
- Most Loki methods currently are synchronous. This issue is not about changing that part of the API. Find, insert, update and remove would stay as they are.
- Only parts of the API that currently use callbacks for **singular** async events would be changed. So the API around event emitters would also remain unchanged.

**These methods / functions would be changed:**
- `Loki` & `Loki.prototype.configureOptions`: The `options` parameter of those has an `autoloadCallback` property, that would not fit a Promise based API. I'm not entirely sure what's the best way to handle this, since simply returning a Promise is not an option for the `Loki` constructor.
- `Loki.prototype.close`: Should be easy to port.
- `Loki.prototype.loadDatabase`: Should be easy to port.
- `Loki.prototype.saveDatabase`: Should be easy to port.
- `Loki.prototype.deleteDatabase`: Should be easy to port.
- `Loki`*`{Fs|LocalStorage|...}`*`Adapter.prototype.loadDatabase`: Should be easy to port.
- `Loki`*`{Fs|LocalStorage|...}`*`Adapter.prototype.saveDatabase`: Should be easy to port.
- `Loki`*`{Fs|LocalStorage|...}`*`Adapter.prototype.deleteDatabase`: Should be easy to port.
- `Loki.prototype.autosaveEnable`: Currently takes an `options` and a `callback` parameter. Since the callback is called after every autosave it is actually an event handler and should not be ported. To prevent confusion, maybe the `callback` arg should be moved into the `options` object.
- `Collection.prototype.ensureIdAsync`: While I'm not completely sure why an async version of the `ensureId` method exists (since it just calls `ensureId` and thus does not prevent lockup of the main thread), it should be easily portable as well.
- `Collection.prototype.async`: This method is only used by `Collection.prototype.ensureIdAsync`. Porting it should be easy but again I'm not really sure why it is part of the Loki API. Maybe it is better to simply remove it?

I only did a quick skim through the code so it is very possible that I missed a method...

As discussed in #474, this (depending on the specific implementation) might be a breaking change and should thus not be landed until 2.0.

Cortys:
Fixes #482.

Since this is quite a big change I want to make sure @techfort and @obeliskos agree with this direction before merging.


obeliskos:
Looks great to me... i'll give @techfort a few hours to voice any issues or merge this (which is on 2.0.0 branch).  

I don't see a problem with the promise polyfill dependency since its dev dependency only used for unit testing.   Similar to how we removed nativescript code (your comment cleanup), eventually i'm guess filesystem adapter will go the same route (external module included with lokijs)

obeliskos:
@techfort might not be around, but I'm sure he would have no issues with this implementation.

So (from your other issue) did he grant you contributor?  If so I would say go ahead and merge... if not let me know and I will merge.

Cortys:
Yes he did. For breaking changes I think it's best if I continue using pull requests for now though, to make sure I'm in line with your intentions for this project.

I'll merge this then.

obeliskos:
Ah great you have the power now... sure no problem... I am a little light on my es2016 proficiency, but I will attempt to adapt and offer higher level structuring advise and code reviews wherever possible.  

As I believe you are actually using this that may help with the fact that we currently have no unit tests for persistence or serialization.  I would imagine we should only need to unit test serialization rather than adapters, and not go through the new 'promisified' loadDatabase and saveDatabase, but feel free to unit test if there is anything that can be done or mocked to ensure others (me) don't break anything.

apporc:
Why this is not reflected in documentation https://rawgit.com/techfort/LokiJS/master/jsdoc/Loki.html. For example db.loadDatabase().then(f)

Viatorus:
Because promise is only available in LokiJS v2:
https://github.com/LokiJS-Forge/LokiJS2

(docs are still missing).

PR #500, opened 2016-10-27
Accept arrays of events

cristyansv:
fix handle of arrays and break the recursion.

### DIFF
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -520,13 +520,14 @@
      * @memberof LokiEventEmitter
      */
     LokiEventEmitter.prototype.on = function (eventName, listener) {
-      var event,
-          self;
+      var event;
+      var self = this;
 
       if (Array.isArray(eventName)) {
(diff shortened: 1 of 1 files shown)


src/lokijs.js
```diff
@@ -520,13 +520,14 @@
      * @memberof LokiEventEmitter
      */
     LokiEventEmitter.prototype.on = function (eventName, listener) {
-      var event,
-          self;
+      var event;
+      var self = this;
```
cristyansv: I think this is important for recursion.

src/lokijs.js
```diff
@@ -520,13 +520,14 @@
...
 
       if (Array.isArray(eventName)) {
         eventName.forEach(function(currentEventName) {
-          self.on(currentEventName, listen);
+          self.on(currentEventName, listener);
```
cristyansv: maybe, word mistake?

techfort:
cheers @cristyansv ! that was definitely some kind of typo.

PR #501, opened 2016-10-27
Return event listener index

cristyansv:
now return a array on event listener index.

### DIFF
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -516,7 +516,7 @@
      * on(eventName, listener) - adds a listener to the queue of callbacks associated to an event
      * @param {string|string[]} eventName - the name(s) of the event(s) to listen to
      * @param {function} listener - callback function of listener to attach
-     * @returns {int} the index of the callback in the array of listeners for a particular event
+     * @returns {int[]} the index of the callback in the array of listeners for a particular event
      * @memberof LokiEventEmitter
      */
     LokiEventEmitter.prototype.on = function (eventName, listener) {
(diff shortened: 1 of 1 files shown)


techfort:
@cristyansv  this looks good, but can I ask why the index is wrapped into an array when returned?

cristyansv:
@techfort I thought doing so for consistency

```javascript
coll.on(['insert', 'update', 'delete'], callback);

// return int[], returns each index of each event

coll.on('insert', callback);

// return int[], return the index of an event.
```

I thought it would be more reasonable to always return the same data type.

techfort:
hey @cristyansv i have to think more about this. It definitely represents a breaking change for anybody who's expecting an integer and gets an array.
Upon review I might add this to the 2.0 branch.

cristyansv:
hi @techfort 

> It definitely represents a breaking change for anybody who's expecting an integer and gets an array.

this is the code prior to my first commit
```javascript
LokiEventEmitter.prototype.on = function (eventName, listener) {
      var event,
          self;

          self.on(currentEventName, listen);
        });
      }

      event = this.events[eventName];
      if (!event) {
        event = this.events[eventName] = [];
      }
      event.push(listener);
      return listener;
    };
```
Currently it returns a **callback**, not an integer or the index of the event.

techfort:
@cristyansv ok i got confused there. If you want to amend the PR to return the callback so it doesn't represent a breaking change i'll be happy to merge.

PR #509, opened 2016-11-19
Make saving atomic in fs adapter

ags131:
This writes to a temporary file then calls rename to replace the
original.
Doing this prevents corruption in cases where a program improperly ends
during a write sequence.


obeliskos:
Hey @ags131, I think when writeFile invokes your callback function it must be setting 'this' to something else.  

Can you create a local variable before calling writeFile like : 
```javascript
var self=this;
```
and then when you call fs rename change that to : 
```javascript
self.fs.rename(tmpdbname,dbname,callback);
```

That should fix the TravisCI build and the fs.rename should actually work.

(When fixed) I am guessing this will not be considered a 'breaking' change as it leaves your filesystem in the same state and filesystem operations should be in the same directory so permissions should not be an issue.  I also think it might be useful if the database fails to save that it will leave the old as backup.

ags131:
Fixed, sorry about that, I'm spoiled to arrow functions where 'this' is carried over from the outer scope,

obeliskos:
Thanks

artch:
@obeliskos When this PR is supposed to be released on npm? If there is data corruption involved, then it seems to be a quite critical hotfix.

obeliskos:
Whenever @techfort gets a chance to update this... we are at a good point to do this so maybe he can find time over the holidays.

tedivm:
Well, I will say that having a critical bug that results in corrupted data sitting here for almost 30 days has made me significantly less likely to use this project.

obeliskos:
As I understand it, loki might be doing an async save when user code throws exception exiting the process.  The solution offered merely keeps a backup of the original database until the save is complete and then overwrites original in an atomic fs rename.  So the 'critical' bug would be in user code and this a protection measure.  @ags131 can correct me if that is not correct but I assume his own p.r. satisfied his own concerns that the problem was external to loki and a backup was sufficient.  If there is a reproducible problem internal to loki then please provide a gist.

@techfort has been extremely busy of late but hopefully he can update npm in the next week or so.

ags131:
@obeliskos This does help in that situation, however, the main goal was to prevent corruption during power loss or program termination (Such as Ctrl+C or SIGTERM)
In these cases, writeFile() is interrupted and ends with either a partial JSON or a 0 byte file.
using the temp file prevents this as renames are atomic and performed at the OS level. If a write is interrupted, the rename would never occur thus preserving the original data. Since rename is atomic, it will always result in either success (File renamed) or failure (file not renamed). In the case of failure the original file remains.

I don't know of a way to make a reliable example or test case as its very timing specific, but I have seen corruption happen many times from quite a few users in the screeps community. We have a mod that manually creates the adapter with this patch that has mitigated most instances of corruption. A new npm release would obsolete that mod and help prevent corruption for unsuspecting users.

obeliskos:
Ah ok thanks.. that is screeps mmo game right?  Looks really well done...  I will have to find some more time soon.

If you are listening to any emitted events make sure to try/catch within them, especially db "close".  If you are using autosave make sure your interval is not too low.  If you need more help diagnosing let me know.  

Your p.r. did raise issue (referenced above) of whether we should try catch listeners ourselves but for some events like insert this would kill performance so we are currently opting not to do so.  Not sure if that is related to your problem but that is one area where external code and loki interoperate.

I am unable to update loki npm so that is why I have not updated it myself.

obeliskos:
@techfort has updated lokijs npm package to current (1.4.2).

obeliskos:
@ags131 (and others) another issue/pr.commit (#526 , 613db681d4f85b2a364f65c12f83a8988bc83357) to be aware of in the event that you receive any instability from loki attempting multiple, simultaneous async io operations conflicting with each other : 

A new [throttledSave](https://github.com/techfort/LokiJS/wiki/LokiJS-persistence-and-adapters#high-frequency-save-throttling) loki constructor option has been added that will hopefully alleviate conflicting save requests.  This is the (only?) area where lokijs could potentially have been -contributing- to instability... through lack of coordination of those requests.

We intend to allow users to test this before making the option default so if you think it may of relevance to you, please do try it to see it helps.   This is not yet on npm but once we verify it we can get techfort to update npm.

Issue #510, opened 2016-11-22
Null-check validation for Collection.update happens to late


sondreb:
The null-check validation that occurs inside Collection.prototype.update happens to late, after the result array have been read from. This throws an error something like "cannot update 0 of null".

This code happens before the check:

`oldInternal = arr[0]; // -internal- obj ref  
`

This happens after trying to read array:

`if (!arr) {
          throw new Error('Trying to update a document not in collection.');
        }`

I'm creating a PR for this issue, will link to it shortly.

obeliskos:
thanks for fixing this! closing..

PR #511, opened 2016-11-22
Moves the null check above attempt to read array. Fixes issue #510

sondreb:
Issue: https://github.com/techfort/LokiJS/issues/510


techfort:
@obeliskos sounds good to me - i'll let you glance over it too because i'm always in such a hurry.

obeliskos:
Great, thanks!

PR #512, opened 2016-11-24
Make operator $in available for index search.

### LINKED ISSUE #508: Enable binary search for $in.
Consider the attached code sample:

```javascript
let db = new loki("test");
let users = db.addCollection("users", {
  indices: ['id']
});

for (let i = 0; i < 10000; i++) {
 users.insert({id: i});
}

let searchIn = {
    id: {$in: [40, 50, 60]}
}

let searchOr = {
     '$or': [{id: {'$eq': 40}},
             {id: {'$eq': 50}},
             {id: {'$eq': 60}}]
}

let iterations = 50000;
for (let j = 0; j < 3; j++) {
    console.time('Function in');
    for(let i = 0; i < iterations; i++ ) {
        const results = users.find(searchIn);
    }
    console.timeEnd('Function in') // Takes ~ 4000 ms


    console.time('Function or');
    for(let i = 0; i < iterations; i++ ) {
        const results = users.find(searchOr);
    }
    console.timeEnd('Function or') // Takes ~200 ms
}
```
I think, $in on indices columns should perform similar to the $or / $eq query.

What do you think about it?

Viatorus:
If operator $in is used with an indexed property:
For each value of the $in array, the index search with the operator
$eq is performed and the result segments are merged

Close #508

### DIFF
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -2442,7 +2442,8 @@
               result.push(t[index.values[i]]);
             }
           } else {
-            for (var i = 0, len = seg.length; i < len; i++) {
+            var len;
+            for (i = 0, len = seg.length; i < len; i++) {
               result.push(t[index.values[seg[i]]]);
             }
           }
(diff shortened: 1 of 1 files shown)


obeliskos:
Nice! Thanks... i'll add a few unit tests feel free to add more if you want.  Also just in case you are curious you should be able to run : 

```
npm install
```
(the first time only), and then 
```
npm run test
```

To run unit tests before submit p.r. so you know what travis ci will complain about.

PR #523, opened 2017-01-05
fix clone returnObj

faceair:
clone returnObj should after `this.add`, otherwise can't get `$loki`


obeliskos:
makes sense, thanks!

PR #527, opened 2017-01-11
5MB LocalStorage Capacity, Not 5GB

xioustic:
I sure wish LocalStorage gave us 5GB of storage. 😉 

Apologies if I am not following any contributing rules, just wanted to fix a trivial mistake in the documentation.


obeliskos:
hah good catch, thanks... I updated the wiki page this was derived from

Issue #525, opened 2017-01-08
Major Update - LokiJS 2.x.x


Viatorus:
Hey everyone,

I want to move forward with you to a new major release of LokiJS.
Let us discuss what our goal for 2.x.x should be and how we reach it.

---

Goals for LokiJS 2.x.x:
 - Use/Migrate to ES6 (promise etc.)
 - Split code base
 - Change concept ?
 - Integrate full-text search #513
 - Solve issues: ?

Procedure:
 - Finish last feature and close feature requests for v1.4
 - Make v1.4 "final" as is and just maintainance for bug fixes
 - Merge changes inside master to branches 2.0.0.
 - Create new repository for LokiJS2.

Viatorus:
Here are some thoughts from me:
- @Cortys and some other of you have already integrate ES6 into the working branch 2.0.0. But currently the master is ~20 commits ahead. I am sure, keeping track of these changes between the branches is not to easy and also demotivating by not knowing the goal.

  To overcome this problem I would stop adding features to the current version (1.4.x) and only fix some bugs (according to expenditure and severity). All new and open feature request should be target for 2.x.x.

  - When should we start to stop featuer request for 1.4.x?
  - How long should we maintenance bugs 1.4.x?

- I recommend to split LokiJSs main file into smaller ones if possible. There are currently ~5.5k lines of code. It would be easier to maintenance. But we should still build a non minified file for easy integration into a single html file (avoid many includes).

- There are currently over 100 issues, some are out of date, solved or irrelevant. Almost all are not labeled.

  - We should tidy up all issues by labeling, removing or solving them. ;)
  - We should also use [githubs project](https://help.github.com/articles/tracking-the-progress-of-your-work-with-projects/) to better manage progress/issues/todos.

That is all from me, what do you think? =)

Cortys:
@Viatorus Some time ago I began working on bringing the changes from master into the 2.0 branch but due to a lack of time I did not finish it.

- Because the work required to merge master into 2.0 will only increase over time, locking 1.4 feature-wise seems like a good idea to me.
- I agree that splitting the codebase should be of high priority. IMHO locking down the amount of changes allowed to 1.4 would be a necessity before this can be done though.

For now I think I'll try to finally complete the master -> 2.0 merge within the next few days. That should make it easier to move forwards with the 2.0 release.

obeliskos:
[...]

PR #528, opened 2017-01-12
Merge master into 2.0 branch

Cortys:
As discussed in #525 I now merged master into the 2.0 branch, converting all new callback based methods to "promisified" ones with as little change to the original implementation as possible.

I merged via multiple smaller merge commits to keep them a bit more comprehensible. Since there were quite a lot of conflicts to resolve, it's probably best if someone else checks whether everything still works as intended.

Hope this helps getting the 2.0 development started.


Cortys:
Tests seem to fail on 4.0 and 0.12 due to missing Promise global in the test environment. Will look into this...

src/loki-fs-structured-adapter.js
```diff
@@ -0,0 +1,255 @@
...
+     * Generator for constructing lines for file streaming output of db container or collection.
+     *
+     * @param {object=} options - output format options for use externally to loki
+     * @param {int=} options.partition - can be used to only output an individual collection or db (-1)
+     *
+     * @returns {string|array} A custom, restructured aggregation of independent serializations.
+     * @memberof LokiFsStructuredAdapter
+     */
+    LokiFsStructuredAdapter.prototype.generateDestructured = function*(options) {
+      var idx, sidx;
```
techfort: as we are in an ES6 context, can we tidy this file up a bit? use `const` for all vars that are not reassigned

src/loki-fs-structured-adapter.js
```diff
@@ -0,0 +1,255 @@
...
+        options.partition = -1;
+      }
+
+      // if partition is -1 we will return database container with no data
+      if (options.partition === -1) {
+        // instantiate lightweight clone and remove its collection data
+        dbcopy = this.dbref.copy();
+
+        for(idx=0; idx < dbcopy.collections.length; idx++) {
+          dbcopy.collections[idx].data = [];
```
techfort: use `Object.keys()` here..

src/loki-fs-structured-adapter.js
```diff
@@ -0,0 +1,255 @@
...
+
+        yield dbcopy.serialize({
+          serializationMethod: "normal"
+        });
+
+        return;
+      }
+
+      // 'partitioned' along with 'partition' of 0 or greater is a request for single collection serialization
+      if (options.partition >= 0) {
```
techfort: this condition is redundant, `options.partition` can only be >= -1, if it's not -1 then this block gets executed, am I getting this wrong?

src/loki-fs-structured-adapter.js
```diff
@@ -0,0 +1,255 @@
...
+        var doccount,
+          docidx;
+
+        // dbref collections have all data so work against that
+        doccount = this.dbref.collections[options.partition].data.length;
+
+        for(docidx=0; docidx<doccount; docidx++) {
```
techfort: again, i believe there is a more FP way to loop over this. Only loop with `for` when there is a need to break the loop

src/loki-fs-structured-adapter.js
```diff
@@ -0,0 +1,255 @@
...
+     * @returns {Promise} a Promise that resolves after the database was loaded
+     * @memberof LokiFsStructuredAdapter
+     */
+    LokiFsStructuredAdapter.prototype.loadDatabase = function(dbname)
+    {
+      var instream = fs.createReadStream(dbname);
+      var outstream = new stream();
+      var rl = readline.createInterface(instream, outstream);
+      var self=this;
+
```
techfort: `var`s begone :)

src/loki-fs-structured-adapter.js
```diff
@@ -0,0 +1,255 @@
...
+          // it should single JSON object (a one line file)
+          if (self.dbref === null && line !== "") {
+            self.dbref = JSON.parse(line);
+          }
+        });
+
+        // when that is done, examine its collection array to sequence loading each
+        rl.on('close', function() {
+          if(typeof self.dbref !== 'object')
+            reject(new Error('Database could not be read.'));
```
techfort: brackets around ifs

Viatorus:
@Cortys  Well done! I will look into this.

@techfort I think, Clemens wanted to first include the feature changes to v2.0.0 and promis-ified the very needed. There can be much more be done for ES6.

For FP (function programming) loops I would be careful because javascript does not really optimize such loops. Standard `for (let i = 0; i < array.length; i++)` are faster.

Cortys:
@techfort Thanks!

Regarding ES6 I completely agree and migrating should probably one of the first things to do with 2.0. I stopped myself from already migrating in this PR to keep it a merge instead of a complete refactoring.

Since there are many ways to use ES6 and code style right now already is pretty inconsistent it might be a good idea to just enforce an ES6 code style via a strict ESLint ruleset (enforcing arrow functions, const decrarations, etc...).

Viatorus:
Yes, I think this is a very good starting point for refactoring.

techfort:
Linting would be a good idea but we'd have to limit it to the files that
are written in ES6, which is a bit ugly. As for fp loops yes, we adopted
for loops because of the speed,  but that is really only relevant for
synchronous queries. An fs adapter can sacrifice that level of speed for
more robust code I'd think,  but like for everything else in my life I
don't have extreme feelings about either approach.

On Thu 12 Jan 2017, 21:09 Viatorus <notifications@github.com> wrote:

Yes, I think this is a very good starting point for refactoring.

—
You are receiving this because you were mentioned.


Reply to this email directly, view it on GitHub
<https://github.com/techfort/LokiJS/pull/528#issuecomment-272284289>, or mute
the thread

Cortys:
@techfort Which files do you have in mind that would not be written in ES6 after migration?

Regarding for-loops I have no strong feelings either.

obeliskos:
Excellent, thanks @Cortys! Feel free to change anything you want.  Maybe our dev dependencies should be upgraded to latest versions and old node versions can probably go.   If this cleanup needs to occur over multiple commits I would be fine with that... better to have good starting dependencies that struggle with old dependencies.   PhantomJS (headless webkit?), karma, jasmine, jshint, (uglify?), etc we will probably want.  Maybe node 6 (or latest LTS 6.9.4?) minimum for travis ci?  

Thanks again... @techfort, once this is upgrade is finalized will we promote it to its own GitHub repo as well?

techfort:
@Cortys @Viatorus @obeliskos first of all a massive thank you - i haven't had time to dedicate to Loki in a long while and you made a fantastic job of it. I approved the PR because i realised that it would be much better to lay out a roadmap of full migration to ES2015 instead of breaking people's balls to tidy up a small part of the lib written in ES2015. Good work everybody, and @obeliskos yes we should promote this to its own repo. Is there a native git way to turn a branch into a repo?

VladimirTechMan:
@techfort @obeliskos (cc @Cortys @Viatorus) I guess it should be the same as (or similar to) this suggested workflow:
https://stackoverflow.com/questions/9527999/how-do-i-create-a-new-github-repo-from-a-branch-in-an-existing-repo/9529847#9529847

(Also, wanted to second @techfort: @Cortys @Viatorus @obeliskos well done, guys!)

obeliskos:
I think your fingerprints are all over that code as well @VladimirTechMan...  

@techfort, please grant all existing contributors (or more) contributor access when you set up new repo,  Vlad's link looks useful whever you get time for that

PR #529, opened 2017-01-13
Added support for throttlesave

mromanbcn:
Batch saveDatabase requests so we don't access disk so often and prevent
an error with async callbacks saving and renaming the exported db.

Note:
I got all tests to pass but one. I need someone to run them and see if they can
spot the issue. The errors come from the test:node execution and from the
test.spec.js file

### DIFF
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -616,7 +616,7 @@
      * @param {adapter} options.adapter - an instance of a loki persistence adapter
      * @param {string} options.serializationMethod - ['normal', 'pretty', 'destructured']
      * @param {string} options.destructureDelimiter - string delimiter used for destructured serialization
-     * @param {boolean} options.saveThrottled - if true, it batches multiple calls to to saveDatabase reducing number of disk I/O operations
+     * @param {boolean} options.throttledSaves - if true, it batches multiple calls to to saveDatabase reducing number of disk I/O operations
                                                 and guaranteeing proper serialization of the calls. Default value is false.
      */
     function Loki(filename, options) {
@@ -633,7 +633,7 @@
       this.autosave = false;
       this.autosaveInterval = 5000;
       this.autosaveHandle = null;
-      this.saveThrottled = false;
+      this.throttledSaves = false;
 
       this.options = {};
 
@@ -650,10 +650,10 @@
       this.persistenceAdapter = null;
 
       // flags used to throttle saves
-      this.saveQueued = false;
-      this.saveRequested = false;
-      this.pendingRequests = [];
-      this.pendingSaveRequestedCallbacks = [];
+      this.throttledSavePending = false;
+      this.throttledSaveRequested = false;
+      this.throttledTierOneCallbacks = [];
+      this.throttleTierTwoCallbacks = [];
 
       // enable console output if verbose flag is set (disabled by default)
       this.verbose = options && options.hasOwnProperty('verbose') ? options.verbose : false;
@@ -814,8 +814,8 @@
           }
         }
 
-        if (this.options.hasOwnProperty('saveThrottled')) {
-          this.saveThrottled = this.options.saveThrottled;
+        if (this.options.hasOwnProperty('throttledSaves')) {
+          this.throttledSaves = this.options.throttledSaves;
         }
       } // end of options processing
 
@@ -1000,10 +1000,10 @@
       case 'constraints':
       case 'ttl':
         return null;
-      case 'saveQueued':
-      case 'saveRequested':
-      case 'pendingRequests':
-      case 'pendingSaveRequestedCallbacks':
+      case 'throttledSavePending':
+      case 'throttledSaveRequested':
+      case 'throttledTierOneCallbacks':
+      case 'throttleTierTwoCallbacks':
         return undefined;        
       default:
         return value;
@@ -1425,8 +1425,8 @@
       }
 
       // restore save throttled boolean only if not defined in options
-      if (dbObject.hasOwnProperty('saveThrottled') && options && !options.hasOwnProperty('saveThrottled')) {
-        this.saveThrottled = dbObject.saveThrottled;
+      if (dbObject.hasOwnProperty('throttledSaves') && options && !options.hasOwnProperty('throttledSaves')) {
+        this.throttledSaves = dbObject.throttledSaves;
       }
 
       this.collections = [];
@@ -2228,62 +2228,54 @@
      * @memberof Loki
      */
     Loki.prototype.saveDatabase = function (callback) {
-      if (!this.saveThrottled) {
+      if (!this.throttledSaves) {
         this.saveDatabaseInternal(callback);
         return;
       }
 
-      if (this.saveQueued) {
-        this.saveRequested = true;
+      if (this.throttledSavePending) {
+        this.throttledSaveRequested = true;
         if (callback) {
-          this.pendingSaveRequestedCallbacks.push(callback);
+          this.throttleTierTwoCallbacks.push(callback);
         }
         return;
       }
 
-      this.pendingRequests = this.pendingSaveRequestedCallbacks;
+      this.throttledTierOneCallbacks = this.throttleTierTwoCallbacks;
 
-      // Store the callback only if it is not undefined and we had requests
-      // coming while we were saving the database. The second condition (after &&)
-      // makes sure we don't include the callback for the call we make internally 
-      // when we detect there were requests while saving the DB. Otherwise the caller
-      // gets (calls to saveDatabase + 1) callbacks.
-      if (callback && this.pendingSaveRequestedCallbacks.length === 0) {
-        this.pendingRequests.push(callback);
+      if (callback) {
+        this.throttledTierOneCallbacks.push(callback);
       }
-      this.pendingSaveRequestedCallbacks = [];
-      this.saveQueued = true;
+      this.throttleTierTwoCallbacks = [];
+      this.throttledSavePending = true;
       var self = this;
 
-      var cb = function() {
+      setTimeout(function() {
         self.saveDatabaseInternal(function(err) {
-
-          self.saveQueued = false;
-          if(self.saveRequested) {
-            self.saveRequested = false;
-            self.pendingRequests.forEach(function(pcb) {
+          self.throttledSavePending = false;
+          if(self.throttledSaveRequested) {
+            self.throttledSaveRequested = false;
+            self.throttledTierOneCallbacks.forEach(function(pcb) {
               // Queue the callbacks so we first finish this method execution
-              // and invoke self.pendingRequests = [] before invoke the callbacks
+              // and invoke self.throttledTierOneCallbacks = [] before invoke the callbacks
               setTimeout(function() {
                 pcb(err);
               }, 1);
             });
-            self.pendingRequests = [];
-            self.saveDatabase(callback);
+            self.throttledTierOneCallbacks = [];
+            self.saveDatabase();
           } else {
-            self.pendingRequests.forEach(function(pcb) {
+            self.throttledTierOneCallbacks.forEach(function(pcb) {
               // Queue the callbacks so we first finish this method execution
-              // and invoke self.pendingRequests = [] before invoke the callbacks
+              // and invoke self.throttledTierOneCallbacks = [] before invoke the callbacks
               setTimeout(function() {
                 pcb(err);
               }, 1);
             });
-            self.pendingRequests = [];
+            self.throttledTierOneCallbacks = [];
           }
         });
-      };
-
-      setTimeout(cb, 1);
+      }, 1);
     };
 
     // alias
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -2236,7 +2236,12 @@
 
       this.pendingRequests = this.pendingSaveRequestedCallbacks;
 
-      if (callback && this.pendingSaveRequestedCallbacks.length > 0) {
+      // Store the callback only if it is not undefined and we had requests
+      // coming while we were saving the database. The second condition (after &&)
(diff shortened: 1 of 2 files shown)


obeliskos:
I think I have way to progress this... but since this area is extremely critical path I think we should have Loki config variable saveThrottled (bool) and default to false at least until it can be determined to not be a breaking change.  

I think the cleanest way to implement this is to keep this logic separate from savedatabase.  So assume we rename the old saveDatabase to saveDatabaseAction (or something) and make something like this wrapper the new saveDatabase() : 
```javascript
    Loki.prototype.saveDatabase = function (callback) {
      var self = this;
      //if(!this.saveThrottled) {
      //  self.saveDatabaseAction(callback);
      //  return;
      //}

      var cFun = callback || function (err) {
        if (err) {
          throw err;
        }
        return;
      };
      
      if (this.saveQueued) {
        this.saveRequested = true;
        setTimeout(cFun, 1);
        return;
      }

      this.saveQueued = true;

      var cb = function () {
        self.saveDatabaseAction(function() {
          self.saveQueued = false;
          if (self.saveRequested) {
            self.saveRequested = false;
            self.saveDatabaseAction(callback);
          }
        });
      };
		
      setTimeout(cb, 1);
    };
```

Please verify the logic is what you were intending.   So one issue was not calling any callbacks when queued... by async calling their callback if queued we are short circuiting an error path so that is why this needs to be opt-in with 'saveThrottled' option for now.  We could implement an array of callbacks to push into and fire/drain that once save is complete but at a minimum while testing lets start with that.

mromanbcn:
Good catch with the missing callbacks.

I followed your suggestion and made 1 change.

Instead of calling setTimeout(callback,1) inside if(this.saveQueued), I added the arrays you suggested and made sure all callbacks are called. If I keep the setTimeout(callback, 1) instead, then I would get a callback before the DB has been saved.

New flag added: saveThrottled, which controls the behavior of save. I modified my local tests and can verify that if I set the flag to false, I get an error and if I set it to true I don't.

The good news are that even if I force throttleSave during the tests, the tests are all passing now. Looks like the callbacks were important.

src/lokijs.js
```diff
@@ -804,6 +811,10 @@
             this.autosaveEnable();
           }
         }
+
+        if (this.options.hasOwnProperty('saveThrottled') && this.options.saveThrottled) {
```
Viatorus: Why the second check? If saveThrottled is false in options, just set it to this.saveThrottled. No need for another check.

obeliskos:
Great, that should just about do it!   We should bring the comment over from savedatabase and maybe mention the database option.

@Viatorus : that line is similar to many other options checks within that (loki.configureOptions) method... it can probably be cleaned up later or on 2.0.0 as a whole but should be fine for now since code path is not performance sensitive.

mromanbcn:
@obeliskos Sounds good, thanks. Just a quick clarification: what do you mean by bringing over the comment from saveDatabase and mention the db option? Is there any file where I can include the information? Thanks

BTW, I will definitely check the loadJSON object and inflate the property automatically. Otherwise, it will be too easy to miss.

@Viatorus: I more or less copied that line from other lines around, but you are right, the second check is not required.

obeliskos:
Well most of our prototype methods have block comments for jsdoc generation defined at their header.  Since we want our jsdoc documentation to document the new saveDatabase instead of the old saveDatabaseInternal, we should : 
- copy (duplicate) the comment declaration currently located before saveDatabaseInternal to apply to saveDatabase.
- remove @memberof attribute for the saveDatabaseInternal so it doesn't show up in jsdoc
- possibly add descriptive text about supporting loki.saveThrottled option

mromanbcn:
Got, it. Absolutely! Will make the changes and submit them for final review. Thanks

src/lokijs.js
mromanbcn: Removed second condition

mromanbcn:
I addressed all the comments and also handled the serialization of the saveThrottled variable. If you can, please look at line 1428. This is where I deserialize the variable and I am giving preference to the value provided in options. Just want to make sure this is the right way of doing it

If you are fine with all the changes feel free to merge or I can do the merge as well.

obeliskos:
Great... thanks.  If you could bear with me, I am trying to track down an issue...  not sure if the issue is my code so I will look over that later today.

obeliskos:
Sorry its taking a while I have not had as much time as I hoped.  I think I have found the problem but I still need to think though the best solution.  

In case you are curious, here is the example that did not work for me and caused me to take pause.  If you load loki sandbox from file system (its in examples directory, use whatever web browser you want) and paste in this sandbox example : 
https://gist.githubusercontent.com/obeliskos/15c1aa87da16cd89b328eb84bbcdf8fa/raw/fe09f7127126e377ea9aeb903574c9fe86eac264/PartitionedMemoryAdapterTestThrottled

...you will see that only the first callback actually fires.   We will need to look carefully at the code path to ensure all possibilities are accounted for and possibly rename some variables and options so they are more intuitive.

I hope to be able to look at this tonight but definitely by the weekend I should be able to help work towards a more final implementation.

mromanbcn:
No rush, lets make sure this works properly. I will try to look at it today and see if I can spot the issue. Thanks!

mromanbcn:
I found the issue. We need to queue the callback execution so we make sure we finish the current method first. The problem was that I was calling the callback and then I was resetting an internal variable. As a result, when you call saveDatabase from the callback, the variable has not been updated yet. By putting the requests in the event loop queue, we make sure we setup the internal state properly.

src/lokijs.js
```diff
@@ -2228,11 +2228,14 @@
      * @memberof Loki
      */
     Loki.prototype.saveDatabase = function (callback) {
+      console.log('1');
```
Viatorus: Remove debugging  purpose.

mromanbcn: Done!

obeliskos:
Thanks, the variable names are still somewhat confusing to me maybe we could rename options/variables to the following (let me know if I misunderstand the purpose of them) : 

options/flags :
- loki.saveThrottled -> loki.throttledSaves
- loki.saveRequested -> throttledSaveRequested (do we need this? maybe check queue length?)
- loki.saveQueued -> throttledSavePending

callback queues (suggestions) : 
loki.pendingRequests -> throttledSaveCallbacks
loki.pendingSaveRequestedCallbacks -> throttledTieredCallbacks

So I guess the logic we are settling on is a two tier callback queue, we add to first tier until such time as our original setTimeout kicks in and from then until that saveInternal finishes we add all future requests to 2nd tier callback queue.  

Thanks, when this is done this is looking like it will be great addition 👍 

Initial phase (setTimeout not yet invoked) (throttledSavePending===false) : 
- add to first tier queue
- if (throttledSavePending===false) setTimeout to queue saveDatabaseInternal

Tiered callback (a save is in progress and has not finished yet) throttledSavePending === true : 
- add to second tier queue

setTimeout callback, saveDatabaseInternal callback : 
- if tier2 queue is not empty, move tier2 queue to tier1, clear tier2, and possibly reinvoke saveDatabase with empty callback param to initiate follow up save

I know I started out suggesting we set up the 'var cb = function()' outer callback but we don't need it (I was thinking it may be factory function).  So can we just inline that?

I should have time to help out over the weekend if needed, and if above isn't clear let me know what to clarify.  My main goal is to simplify and distill this algorithm and reduce edge cases wherever possible... since it is not performance sensitive we can code for clarity.

mromanbcn:
Thanks @obeliskos for the comments.

I agree with your name changes, but let me suggest one change for the callbacks queues:
1. loki.pendingRequests -> throttledTierOneCallbacks
2. loki.pendingSaveRequestedCallbacks -> throttledTierTwoCallbacks

I think this makes the code easier to read as it is clear we use a tiered solution and you can easily spot tier one and two queues.

Regarding throttledSaveRequested, I thought also we could remove it but after going over the code I feel it would make it more difficult to read and more error prone. The queue length would have a semantic meaning. I would prefer to make it explicit and keep the flag. Let me know if you agree. Also,I found that if the caller does not pass a callback (undefined), the queue length would be 0 but we still need to remember there were pending requests.

Your algorithm description is correct. One minor comment:

<Initial phase (setTimeout not yet invoked) (throttledSavePending===false) :

<1. add to first tier queue
<2. if (tier1 queue length === 1) setTimeout to queue saveDatabaseInternal

2 is not strictly true. Regardless of the queue length we always call saveDatabaseInternal. I think you got that condition from my original code where I was deciding whether or not to queue the callback. The good news is that I was able to remove that condition, so the code should be more clear. The reason I was able to remove it is because when I internally call saveDatabase again because there were pendingRequests, I do not pass the original callback anymore. You suggested this in your comment and you were correct. Passing the callback again was triggering an extra callback invocation. No need to pass it, because the callback has already been invoked when in the initial pass we stored it in the tier one queue and then invoked it before calling saveDatabase again

Finally, I also inlined the method.

I am updating the pull request now, let me know if you agree, otherwise I will be happy to work on it.

obeliskos:
So I've had time to think about this and I think I need to amend the approach I outlined above.   I am trying to juggle all the different possibilities this could break and I think we accommodate situation if they don't yield thread... not sure that would be pretty code but i'd like to not wait to initiate I/O which -is- truly async.

So I took our algorithm and reorganized it and I think it looks good but I will let you weigh in.  I am open to suggestions but due to rework, I got rid of 1 of the loki level arrays since I will use local variable to snapshot it.  Whenever you or anyone gets a chance, can you look over this logic to see if it will work for all cases you can imagine? 

```javascript
    Loki.prototype.saveDatabase = function (callback) {
      var self=this;
      var localCallbacks;

      // if not using throttled saves, invoke save and return
      if (!this.throttledSaves) {
        this.saveDatabaseInternal(callback);
        return;
      }

      // if there is a save in progress, tier and throttle this for follow up save
      if (this.throttledSavePending === true) {
        this.throttledCallbacks.push(callback);
        return;
      }

      // copy to local var, clear loki callbacks, and flag save pending
      localCallbacks = this.throttledCallbacks;
      this.throttledCallbacks = [];
      // add our own callback first
      localCallbacks.unshift(callback);
      // flag pending
      this.throttledSavePending = true;

      // do an actual save
      self.saveDatabaseInternal(function(err) {
        // callback means i/o is done
        self.throttledSavePending = false;

        // invoke each local callback
        localCallbacks.forEach(function(pcb) {
          // Queue the callbacks so we first finish this method execution
          if (typeof pbc === 'function') {
            setTimeout(function() {
              pcb(err);
            }, 1);
          }
        }); 

        // since this is called async, future requests may have come in, if so.. kick off next save
        if (self.throttledCallbacks.length > 0) {
          self.saveDatabase();
        }
      });
    };
```

mromanbcn:
I modified the code in the pull request to address your latest suggestions. Everything seems to be working fine with the existing tests and the scenarios I have been testing as well.

Based on the latest changes, I even tried to remove the setTimeout around the callback invocation and worked as well. However, I decided to leave it. I think it is safer to delay the callbacks until we finalize the method execution so the DB code is not affected by strange behaviors in the callback handlers. If you prefer to remove it feel free to edit the pull request.

Viatorus:
@mromanbcn Great work! Thank you very much! :)
> the scenarios I have been testing as well

Feel free to add your test cases as well. ;)

@obeliskos Would it be possible to port this to v2.0.0? ;)

obeliskos:
Yes, thanks @mromanbcn !  I will look into unit tests (unless you want to)... I need to either make sure our jasmine tests work with async calls or if by removing many of our setTimeouts they will work enough to detect fail conditions.  The issue I ran into earlier that I posted a gist for has a unit test almost exactly like it but the 'expect' statements were not evaluated.  May be some settings we can tweak.

mromanbcn:
Thanks @obeliskos and @Viatorus 

Let me ask you a favor. I am going to copy here the code I have been using to test the error. Would you mind adding it to your unit test framework? If I make more contributions I will definitely add tests myself in the future. I am a bit short on time right now

If you run the code below without throttledSaves:true then you will see an error from the DB complaining that the db~ filename is missing. That's where the async behavior does not allow preserving the ordering of the calls. As soon as you add the throttledSaves:true, the error goes away and things are properly saved. Also, you can check that the number of invocations matches the number of callbacks.

```js
var loki = require('lokijs');
var db = new loki('loki.json', {throttledSaves:true, autoload:true, autoloadCallback:go});

var s = 0;
var t = 0;

var children;

function insert() {
	console.log('Inserting: ' + s);
	children.insert({name:'' + s, legs: 8});
	s++;
  	db.saveDatabase((err)=>{
  		console.log('Callback: ' + t);
  		t++;
  		console.log(err);
  	});
}

function go() {
        children = db.getCollection('children');
	if (children === null) {
		children = db.addCollection('children');
	}
	const max = 10000;
	var i = 0;
	for(i; i < max; i++) {
		setTimeout(insert, 50);
	}
}
```

mromanbcn:
One final thing...

It would be great if we could turn the throttled behavior into the standard as soon as we can get more people to use it and confirm they don't see problems. While the DB is synchronous, the save method is not (because of the 2 calls to save and rename files) and therefore the results are unpredictable. With the new changes we guarantee predictable results. This gets very tricky especially in cases where we call saveDatabase and then immediately call loadDatabase (I found this behavior in a couple of tests I modified). There are no guarantees that load will pick up the new changes. Even with the new changes and putting loadDatabase in the saveDatabase callback, you are not guaranteed to pick up the new changes unless the user sets the throttledSaves flag to true.

Someone commented on my original issue mentioning he had had the same problem, so this would avoid the problem for new users invoking saveDatabase themselves

Thanks again and looking forward to using 2.0.0!

obeliskos:
Yep I can look at adding something for that.  Thanks again!

PR #554, opened 2017-04-04
Update v2

Viatorus:
Hey guys,

I just aligned LokiJS v1 and v2 with bugfixes and "features" to don't lose track.

I did my best for throttled save. Would be cool if someone can look over this.
I am quite pleased with the saveDatabase promise handling.


obeliskos:
not sure if/when @techfort will get around to reviewing so feel free to commit this whenever you want @Viatorus

PR #562, opened 2017-05-03
Update .npmignore

segphault:
Updating .npmignore, should bring package install size down from 5.3MB to 528K.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,5 +1,8 @@
 docs/
-benchmarks/
+benchmark/
 demos/
 examples/
-jsdocs/
+jsdoc/
+tutorials/
+spec/
+coverage/


Viatorus:
Does somebody know, why this pull request failed? I don't get it.

obeliskos:
I haven't had time to look into this and other issues while working on binary indices, but we do have a few 'async' tests (like in persistence.spec.js where the 'it' is passed a 'done' variable).   Not sure if its a problem with the test or with changes to phantomjs or mocha since our devdependencies use [carets in their versioning](http://stackoverflow.com/a/22345808)... could be why the tests still work for me but fail on some installs.  I think i encountered this myself on a fresh ubuntu install i had done not long ago.

Otherwise I am in agreement with his suggested changes except maybe keep tutorials folder so that full jsdocs can be generated with **npm run jsdoc** if they later decide to do so.

obeliskos:
So this may have been bad unit test that stopped working as minor dependency was released.  I fixed the unit test so this passes now... thanks.

Issue #578, opened 2017-06-26
Optional delta style object for update operations in changes system


ChristopherTotty:
Currently when using the changes system an update is stored as a complete copy of the object.  It would be beneficial in some cases to instead store a delta object with only the properties that have been changed from the original.

This would be most helpful when using LokiJS as a persistence layer for client data that will be synced to a central database on a server, since it would reduce the amount of traffic that is sent as part of that sync process.

I am currently working on a project where this is the case, and I am considering making this change for my own use.

Would there be any interest in a pull request for this feature, and if so, is there any opinion on it before I get started?

obeliskos:
Sure, that would be great 👍 

By default, collections do not have cloning enabled so all returned objects are the same instances as those used internally.  If I do a find and update the object(s), the internal references are already updated before I call the update() method which essentially just rebuilds indices and emits events.

So if cloning is not enabled, I guess you would have to 'pre-emptively' add cloned baseline copies of all documents when initializing changes api on collections that opt in.   Then I guess you would not include those in changes, but upon generating changes you would need to rebaseline 'non-dirty' baseline objects.  With those copies, you can compare each updated object with highest (most recent) history object (or if empty) the baseline object.  Does that make sense?  

@techfort wrote changes api so if he can think of any issues maybe he can chime in.

ChristopherTotty:
Ah, I didn't think about cloning since we added our own implementation that is run on everything as it is pulled out of the database.  

Seems like doubling the memory consumption of every collection that opts in would be a turn off.  

What would you think about either requiring cloning to be enabled or potentially building that clone set as items are selected from the collection?  That way memory use would start at the same level as the current implementation and slowly grow to potential max of double as items are pulled from the collection.

obeliskos:
So let's say you add an collection constructor option... something like 'deltaChangesApi' which you can turn on.  If they turn that on I can see us forcing 'clones' collection option to be 'true' whether they pass it or not...  possibly even override cloneMethod to ensure we use 'parse-stringify' instead of 'shallow' cloning methods which may not be effective at tracking changes to nested objects.  

Since you are introducing a new option, as long as the old behavior continues to work as it did before, you can override the other options as it makes sense to do so.  Does that answer your concerns?  

With cloning enabled there is still (at least temporary) memory overhead and performance hit to clone internal objects as they come in and go out through find/insert/update/etc ops.   I can see that being perfectly fine though as your feature would be concerned with reducing transfer size of changesets.

[...]

PR #582, opened 2017-07-07
#578: Add optional flag for Changes API to limit update operation out…

ChristopherTotty:
#578: Add optional flag for Changes API to limit update operation output to modified properties only.  Also includes properties flagged as unique, since they would be needed to identify the item that is updated.

### DIFF
--- a/spec/generic/changesApi.spec.js
+++ b/spec/generic/changesApi.spec.js
@@ -54,4 +54,43 @@ describe('changesApi', function () {
 
     expect(users.getChanges().length).toEqual(0);
   });
+
+  it('works with delta mode', function () {
+    var db = new loki(),
+    options = {
+      asyncListeners: false,
+      disableChangesApi: false,
+      disableDeltaChangesApi: false
+    },
+    items = db.addCollection('items', options );
+
+    // Add some documents to the collection
+    items.insert({ name : 'mjolnir', owner: 'thor', maker: { name: 'dwarves', count: 1 } });
+    items.insert({ name : 'gungnir', owner: 'odin', maker: { name: 'elves', count: 1 } });
+    items.insert({ name : 'tyrfing', owner: 'Svafrlami', maker: { name: 'dwarves', count: 1 } });
+    items.insert({ name : 'draupnir', owner: 'odin', maker: { name: 'elves', count: 1 } });
+
+    // Find and update an existing document
+    var tyrfing = items.findOne({'name': 'tyrfing'});
+    tyrfing.owner = 'arngrim';
+    items.update(tyrfing);
+    tyrfing.maker.count = 4;
+    items.update(tyrfing);
+
+    var changes = db.serializeChanges(['items']);
+    changes = JSON.parse(changes);
+    
+    expect(changes.length).toEqual(6);
+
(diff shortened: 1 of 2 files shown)


Viatorus:
Looks good to me. Could you please add a unit test?

obeliskos:
Do you want this setting to be persisted across database saves/loads?  In our current 'loadJSONObject' is where we 're-wire' up the collections on load and since changesAPI wires up listeners we ['rewire' changes api options via constructor options](https://github.com/techfort/LokiJS/blob/7dd2f65295e7b800378d7e02cfeedd528290b217/src/lokijs.js#L1540).  So to restore a saved collection with changes data being interpreted properly you might want to add your option to that constructor.

Might be good to see a unit test as @Viatorus had suggested, looks like we have simple unit tests for changesApi in /spec/generic/changesApi.spec.js so it should be pretty easy to add there... let me know if you need any help with that.  I see @techfort set asyncListeners to false which might make unit testing easier for you as well.

obeliskos:
Also I was trying to test nested objects and i'm not sure it picked up my change, can you load this within the /examples/sandbox/LokiSandbox.htm from your filesystem (so that it uses your changes to lokijs.js) : 
```javascript
var db = new loki('sandbox.db');

// Add a collection to the database
var items = db.addCollection('items', { disableChangesApi: false, disableDeltaChangesApi: false } );

// Add some documents to the collection
items.insert({ name : 'mjolnir', owner: 'thor', maker: { name: 'dwarves', count: 1 } });
items.insert({ name : 'gungnir', owner: 'odin', maker: { name: 'elves', count: 1 } });
items.insert({ name : 'tyrfing', owner: 'Svafrlami', maker: { name: 'dwarves', count: 1 } });
items.insert({ name : 'draupnir', owner: 'odin', maker: { name: 'elves', count: 1 } });

// Find and update an existing document
var tyrfing = items.findOne({'name': 'tyrfing'});
tyrfing.owner = 'arngrim';
items.update(tyrfing);
tyrfing.maker.count = 4;
items.update(tyrfing);

var result = db.serializeChanges(['items']);
result = JSON.parse(result);
logObject(result);

inspectObject(db);
```

It seems as though your getObjectDelta() is recursive to handle nested objects?  Not sure why i get empty object on first update (when i did not change object) and empty object on second update (when i did change a count)... how are nested object property changes intended to be supported.  I suppose this will need recursive custom 'reapply' logic to object assign at each existing subobject?

ChristopherTotty:
Thanks for your comments. I will look at the config and nested objects when I get back in the office Monday. One question, I know this may sound dense, but what is the best way to run the unit tests locally for me to verify my new test?

obeliskos:
if you have done an : 
```
npm install
```
you can then run : 
```
npm run test
```

ChristopherTotty:
Sorry last week was a busy one, so I was only able to get to back to this today.  Please let me know if you have any other thoughts.

ChristopherTotty:
Ah, sorry, this change did not address setting the config values.  I'll get another commit in for that soon.

obeliskos:
'does what it says on the tin' 😏 

looks great... thanks!

PR #584, opened 2017-07-12
Fixed asynchronous deleteDatabase

foxel:
Fixed asynchronous deleteDatabase to actually wait for database deletion query complete before executing callback.


obeliskos:
Thanks, looks good...

PR #585, opened 2017-07-13
Fix require statement

ozcanovunc:
loki-indexed-adapter.js is missing under build directory, renamed as loki-indexed-adapter.min.js


obeliskos:
can you explain exactly what this p.r. does?

ozcanovunc:
`adapter=require("./loki-indexed-adapter.js")` was failing because under **build** directory there is `loki-indexed-adapter.min.js` file instead.

I just renamed loki-indexed-adapter.js as loki-indexed-adapter.min.js.

Kanaye:
Correct me if I'm wrong but isn't this fix just 'temporary' until a new version is build ?
You 've just changed the minified file and not the source so this change will probably never make it into a release.

ozcanovunc:
I don't know how you generate the minified code. So this is just a workaround. You guys can fix it properly later on knowing this problem :)

obeliskos:
Ah, I think I see (more) what you were trying to do...  

So what is giving you the error?  Is it something in your build toolchain like browserify or are you actually making a javascript call to [getIndexedAdapter](https://github.com/techfort/LokiJS/blob/master/src/lokijs.js#L830-L840)?  

The only reason that method exists in the first place was to try to allow bundlers like browserify to automatically bundle the indexed adapter.  I am inclined to want to just delete that function as it is kind of a hack for not having our indexed adapter in its own repo, but i am not sure if anyone is using this in their build chain.

So are you 'requiring' lokijs as a defined project dependency or an absolute filename to the minified lokijs?   I would imagine a require("lokijs") would resolve to the unminified version in the src folder while a require of some relative path the min file would expect unminified which is not in build.

Most times, to do the change I think you were trying to do, you would modify the src/lokijs.js file and modify the function (highlighted if you click on the link above).  Then you would save that and run : 
```
npm run build
```

Which should run unit tests and then regenerate minified files.  Then you would check in changes to both lokijs.js and the regenerated min file(s).

Let me know how you intend to use this function (toolchain or calling directly via javascript).

Kanaye:
Webpack can parse and require it without problems.
I've used webpack successfully with lokijs.
But it wouldn't be a problem to remove the function `getIndexedAdapter`.
If you require lokijs via webpack you can also just `var IndexedAdapter = require('lokijs/src/loki-indexed-adapter.js')` and pass it to the Loki constructor.

obeliskos:
Thanks Kanaye, I wonder if it would be better to use your syntax in that function, as that would likely resolve correctly for most cases (rather than the relative path)?

ozcanovunc:
In our project, for few cases we had to require our modules with relative path. So as a habit I used lokijs in the same way.

But using `require("lokijs");` fixed my problem. Everything runs correctly :)

Kanaye:
Sorry for the late response.
I'm not sure if changing the the syntax is the better option.
It *should* work in most cases with nodejs require style compatible tools like webpack.
But there are some edge cases e.g. when someone clones the repository under some other directory name.
But I'm not sure if we should care about cases like that.

Also I'm not sure if this is really an issue because normaly you won't `require()` minified versions of those files.
In node minification is in most cases irrelevant and tools like webpack compress/uglify the output themself.

PR #586, opened 2017-07-19
Small fix in cloning methods

ichursin:
* add correct saving object prototype when we cloning data

### DIFF
--- a/spec/generic/cloning.spec.js
+++ b/spec/generic/cloning.spec.js
@@ -63,7 +63,31 @@ describe('cloning behavior', function () {
       expect(result.name).toBe("mjolnir");
     });
   });
-  
+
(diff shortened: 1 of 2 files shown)


obeliskos:
I'll try to look at this tonight...  sounds similar to our not very well documented ['protos'](https://github.com/techfort/LokiJS/blob/master/spec/generic/typed.spec.js) functionality.  Do you persist your data and if so are the protos attached correctly across saves?

obeliskos:
This seems to work great with some of the edge cases I threw at it (forceClones on chained data, as well as defining collection protos to inflate with in db constructor).  We should probably cover all those edge cases better, I will look at that if I get a few minutes later...  

Thanks!

PR #608, opened 2017-09-05
Prevent 'meta' namespace collision fixes #368

### LINKED ISSUE #368: Internal key 'meta' is too generic
The internal key `meta` is too generic. It could easily collide with a `meta` key from a user-provided object. I propose to change it to `$loki_meta` or something that has less chances to collide with user data.

What do you think @VladimirTechMan @techfort @obeliskos ?

Redsandro:
`meta` is too generic and causes namespace collisions with many existing input documents. This should be considered a bug and we need to fix this in order for `LokiJS` to be usable with todays API outputs.


Redsandro:
How can the test fail? I think the test must have failed before I made changes.

Redsandro:
I was wrong. Testing on `master` is fine. I don't know what I missed.

carchrae:
i'm not the owner of this project, but i see you're investigating and on the case.  :+1:

obeliskos:
Failure is in memory adapter paging size tests... you probably increased the size of the objects just enough that two of them couldn't fit into one '128byte page'... test.   

I should probably tweak the page size to just ensure only 1 object will fit on a page and raise expect to 5.

Redsandro:
@obeliskos I shouldn't have tried to PR on top of a project that is heavily optimized. :stuck_out_tongue_winking_eye: 

Is this a relatively easy fix I can apply on top?

obeliskos:
no its a ridiculously small page so that i could test paging with only a few test documents but apparently just big enough that more than one serialized document was able to fit with in it before but not after... i'll try to make fix that in a min.

obeliskos:
ok, i forgot to reference this pr or related issue, but i have checked in changes to that unit test so that it would pass (both) before and after pr.  

as mentioned though this would have to be a collection or database level option like : 
```javascript
// either
var coll = db.addCollection("users", { useAlternateMeta: true });
// or
var coll = db.addCollection("users", { alternateMeta: '$lokimeta' });
// or 
var db = new loki('test.db', { useAlternateMeta: true });
// or 
var db = new loki('test.db', { alternateMeta: '$lokimeta' });
```
(option names are just examples) 
i am up for hearing pros and cons, i might go for user provided name prop on collection if it were me but this would have to be repeated everywhere in that case, if at db level you would probably need module global variable to keep setting/value in or inject into collection options in collection constructor.

obeliskos:
Hey @Redsandro , were you planning on revisiting this or should we close this?

Redsandro:
I was planning to revisit, but more pressing matters keep falling on my path.
I will suspend this for now. I might open a new PR sometime in the future, but I encourage anyone to beat me to it.

obeliskos:
Ok, thanks... I certainly understand.   So to summarize as we put this on hold : 

For this particular case where you are inserting documents which may already have conflicting meta properties, we would need a pr like this to allow defining alternate meta property name.

I'll add that for other use cases which are only concerned with our meta properties when extracting from a loki database, we have a ['removeMeta'](https://rawgit.com/techfort/LokiJS/master/jsdoc/Resultset.html#data) option you can pass to chained data() call to remove $loki and meta properties from results.

Redsandro:
Hi @obeliskos, yes you are correct. But also, I wanted to (investigate if Loki works properly when we) stop `meta` from being created in the first place.

I'm having a performance-critical app, and I think in my case the metadata is irrelevant. Every time I insert 10000 documents, Loki will iterate all that and create 10000 new keys on them, with some dates. Does it also create 10000 objects (`meta`)? Or does it make 1 object and assign it's reference to the objects that are inserted?

All I really want is to do everything by reference. Not add objects on insert, and force clones on result data.

This probably breaks database storage adapters, but I'm using Loki with read-only live generated data.

obeliskos:
So we might want to more officially support this later but this can sort of be done now if you are willing to take responsibility over collection (meaning this is currently unsupported and you could mess up collection integrity of you mix this method with other insert/update/remove methods)

I have created a loki sandbox gist for utilizing [standalone collections as 'snapshots'](https://rawgit.com/techfort/LokiJS/master/examples/sandbox/LokiSandbox.htm#rawgist=https://gist.githubusercontent.com/obeliskos/b7bbbb9681739ea44e544c2e25752774/raw/2c5e5659daf58369daf451da9d56dec37e26a9cf/Collection%2520Snapshot) whereby you : 
- manually construct the collection outside of the context of a 'loki' database object instance
- you manually assign the documents to the internal 'data' property
- you manually apply any indices (if needed)
- you perform your queries
- you only use that collection instance again for querying
- the example gist uses detached collections, instances via 'new loki.Collection' which is safer than attempting this within a database instance which you might want to persist.  If you never need persistence you could probably aggregate them within a single database using loki.addCollection method.

Redsandro:
This looks rather interesting @obeliskos, thank you for writing up this gist.

Do I need to create an index manually to immitate what happens when I would insert a collection "by the book"?

obeliskos:
Normally indices are maintained as you insert/update/remove, but since this method circumvents that process, you will need to run ensureIndex to update binary indices after every time you 'set' or modify the 'data' document array.

Normally on insert, we create $loki and meta properties.  The $loki has its own unique index which is leveraged by calls to collection.get().  The method i described adds neither $loki nor meta properties.  

Preferably you would might want to 'bring you own' unique id in your document and apply an unique index on that column.  With a unique index you would perform a [collection 'by()'](https://rawgit.com/techfort/LokiJS/master/jsdoc/Collection.html#by) to lookup by a unique key defined by collection.ensureUniqueIndex(propertyName) invoked after changing data array  : 
```javascript
coll.ensureUniqueIndex("id");
```

So I guess I would verify how you were using the $loki id before if you were not persisting the database. Do you have another uniquely identifiable property you can make unique?

Redsandro:
@obeliskos thank you for your elaborate response.

>  I would verify how you were using the $loki id before

I'm not sure how it was used internally, but my simplified usage is:

* `loki.insert([a bulk of documents])`
  * once
* `loki.chain().find().compoundsort().offset().limit().data().map()`
  * Dozens of times, with non-predictable find and sort parameters

(PS - using `map` because `mapReduce` has a `reduce` stage which I don't need)

> Do you have another uniquely identifiable property you can make unique?

Yes, but they are of type string, in some cases like `ObjectID` from MongoDB.

---

BTW, can I do this to make separate collections?

```
let coll1 = db.addCollection('foo', {clone: false})
let coll2 = db.addCollection('foo', {clone: false})
let coll3 = db.addCollection('foo', {clone: false})
```

Or does `foo` need to be unique because db links them internally?

(An unknown number of collections is made on the fly. Not sure if I need to randomize the collection name)

PR #609, opened 2017-09-05
Just to force CI testing on master

Redsandro:
This can be removed in a minute or 5


Redsandro:
Ok, testing on `master` is a succes. Now I know.

PR #614, opened 2017-09-14
fix vscode jsdoc crash

gregolai:
vscode 1.16.0 crashes when parsing src/lokijs.js. I narrowed it down to this jsdoc here. Tested with fix and it no longer crashes.


obeliskos:
I can't imagine how you narrowed that down, but thanks!  I use vscode but had not encountered that yet.

PR #635, opened 2017-12-04
addCollection returns existing collection

cepm-nate:
addCollection returns existing collection if an existing collection exists with the same name. Original options are not overwritten.

### DIFF
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -1010,6 +1010,15 @@
      * @memberof Loki
      */
     Loki.prototype.addCollection = function (name, options) {
+      var i,
+        len = this.collections.length;
+
+      for (i = 0; i < len; i += 1) {
+        if (this.collections[i].name === name) {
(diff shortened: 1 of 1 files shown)


obeliskos:
Seems some nested dependency on hawk package (probably via karma/phantomjs but [similar to this](https://github.com/hapijs/boom/issues/164)) broke old versions of node... and a few bugged unit tests which your p.r. helped shine a light on broke the 4.0 travis build.

Rather than chase down our caret dep and switch to tilde deps and verify with a travis vm running old version of node, I have updated the travis config to target 4, 6, and 8 versions of node instead of 0.10, 0.12, and 4.0.  I am sure loki will run fine and integrate into older node dependency chains but our versioning with current toolchain should hopefully continue to support node 4.0 indefinitely. 

@techfort or @viatorus let me know if you have ideas for other solutions.

PR #643, opened 2018-01-15
Fixes autosave race condition when using asynchronous adapter

radex:
Consider this scenario:

- autosave is enabled (easiest to see issue with low autosave timeout, like 150ms)
- asynchronous adapter (say, IndexedDB adapter) is used

and the following steps:

- a bunch of changes (1) are performed. DB is now dirty
- Loki checks if DB is dirty. It's dirty
- saveDatabase() is called
- adapter save is scheduled (but it's asynchronous)
- a bunch of more changes (2) are performed. DB is still dirty
- adapter save is finished
- dirty flags are reset
- Loki checks if DB is dirty. It's not dirty

… except it actually is. But if you don't perform any more changes and reload the website, changes performed between the beginning and end of saveDatabase() will be lost.

Since saveDatabaseInternal clears dirty flags anyway (regardless of whether the save was successful or not), I just moved the dirty flag check to happen synchronously with the call. The effect in success/fail is the same, except now my issue is fixed.

But this doesn't work for the reference mode export (partitioning adapter does its own dirty checking). So probably a better long-term solution would be to have a second dirty flag for when save is in progress, or instead of a boolean flag, increment a dirty counter.

### DIFF
--- a/spec/generic/persistence.spec.js
+++ b/spec/generic/persistence.spec.js
@@ -706,6 +706,30 @@ describe('async adapter tests', function() {
     });
   });
 
+  it('verify there is no race condition with dirty-checking', function(done) {
+    var mem = new loki.LokiMemoryAdapter({ asyncResponses: true, asyncTimeout: 50 });
+    var db = new loki('sandbox.db', { adapter: mem });
+
+    var items = db.addCollection('items');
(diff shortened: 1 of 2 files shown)


obeliskos:
Good catch... and thanks for the unit test

Issue #674, opened 2018-05-05
Module not found: Error: Can't resolve 'fs'


gimox:
I know this is a old problem with a response.. .add fs as empty in webpack config.
Ok! but what about if you can't do this like angular cli project (i don't wan't do a angular eject).
So i run my project in a browser and "fs" module is not require. 

My proposal is to make this line safe `this.fs = require('fs');`

obeliskos:
Is this a runtime error or a error such as during scaffolding?  If error occurs at runtime, what persistence adapter do you expect to use (or do you not need to persist database)?

gimox:
it's  during scaffolding, i'm using indexdDB and it working great. So this fix generate a warning but now it  can serve and build with angular cli.

obeliskos:
ok, i think if we had to do it again we would have extracted -all- persistence adapters out into their own modules.

wait, we are doing it all again with LokiDB, and we did :)

obeliskos:
Thanks for pr!

gimox:
👍👍

gdpelican:
Version bump please? :D

obeliskos:
I will work on updating the changelog and ask @techfort to update maybe around the weekend.

If anyone finds any issues before then, post them so we can fix before version bump.

PR #675, opened 2018-05-05
#674 fix for browser project: fs not found

gimox:
fix for angular/cli project with error: Module not found: Error: Can't resolve 'fs' 
It can fix other project with webpack when user can't edit webpack config.


gimox:
Ok thanks

PR #690, opened 2018-06-26
Update lokijs.js

jonathanz:
make transform support $or and $and which top children are Array, current transform use 'shallow-recurse-objects', which not deep clone Array, change to use 'parse-stringify' for it.
Example transform:
    var memberleveltx = [
        {
            type: 'find',
            value: {
                '$or': [
                    {
                        'crowd.anyone': true
                    },
                    {
                        'crowd.specified.grade.value': {$contains: '[%lktxp]grade'}
                    }
                ]
            }
        }
    ];


jonathanz:
close it.

PR #692, opened 2018-06-27
change 'shallow-recurse-objects' clone to support deep Array clone

jonathanz:
fix transform not support $or and $and issue, for $or and $and whose top children are Array, current transform use 'shallow-recurse-objects', which not deep clone Array, change 'shallow-recurse-objects' clone to support deep Array clone.
Example transform:
    var memberleveltx = [
        {
            type: 'find',
            value: {
                '$or': [
                    {
                        'crowd.anyone': true
                    },
                    {
                        'crowd.specified.grade.value': {$contains: '[%lktxp]grade'}
                    }
                ]
            }
        }
    ];


obeliskos:
Hey @jonathanz can you provide a unit test in spec/generic/transforms.spec.js to cover the use case which this fixes?

jonathanz:
Hey @obeliskos uint test for this fix has been added. please check it.

jonathanz:
Hey @obeliskos uint test for this fix has been added. please check it.

obeliskos:
Looks good, thanks for pointing this out and fixing... that would seem to be a somewhat common use case we overlooked.

PR #693, opened 2018-07-01
Reference `process` on the global object

bmac:
Hi, 

When using loki.js in a browser environment I am getting the error `Uncaught ReferenceError: process is not defined` when the browser tries to evaluate the getENV() line since `process` is not a global variable in the browser and [use strict](https://github.com/bmac/LokiJS/blob/721748c86a0392900de081651ad7e87536296a9c/src/lokijs.js#L21) is enabling strict mode at the top of the loki.js file.

### DIFF
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -831,7 +831,7 @@
           return 'NODEJS';
         }
 
-        if (typeof global !== 'undefined' && global.window && process) {
+        if (typeof global !== 'undefined' && global.window && global.process) {
           return 'NODEJS'; //node-webkit
         }
 


obeliskos:
Sorry I have not attended to this promptly, I have been pretty busy to try to figure out what is going on with this...

So if you go to our [Loki Sandbox page](https://rawgit.com/techfort/LokiJS/master/examples/sandbox/LokiSandbox.htm), do you get the above error in your browser console?  I don't see that in Chrome... i do see a random html typo though :( 

Are you using any type of bundler on your code like webpack?

bmac:
I'm using it with ember-cli. I suspect there is a babel plugin somewhere that is injecting `global` into the scope. I'll try to reproduce it again later tonight when I'm in my other laptop.

bmac:
In my project I tracked down `window.global` and determined it is injected by the `babel-polyfill` package.

So it seems like the issue I'm trying to fix here only impacts users with the`babel-polyfill` package. Feel free to close this pull request if you feel this change is out of scope for your project.

obeliskos:
Could you verify if replacing : 
**&& process** 
with 
**&& typeof process !== "undefined"**
accomplishes the same thing without a strict warning?

PR #706, opened 2018-08-28
Corrected $aeq example in Query Examples

mattezell:
When copying/pasting from the $aeq example, I was greeted with a TypeScript error indicating 'Expected 0-1 arguments, but got 2.' - wrapping as a object makes it conform to the PartialModel shape as is expected by find().


obeliskos:
Thanks!

PR #709, opened 2018-09-26
Allow undefined check on nested objects & add $exists operator

plepe:
1. commit fixes this behaviour:
* db.find({ 'someprop': undefined }) would find any object where a someprop is not set
* db.find({ 'nested.someprop': undefined }) would not find any object where someprop on 'nested' is not set

2. commit adds a $exists operator as in MongoDB (description copied from MongoDB):
Syntax: { field: { $exists: &lt;boolean&gt; } }
When &lt;boolean&gt; is true, $exists matches the documents that contain the field, including documents where the field value is null. If &lt;boolean&gt; is false, the query returns only the documents that do not contain the field.

### DIFF
--- a/jsdoc/tutorial-Query Examples.html
+++ b/jsdoc/tutorial-Query Examples.html
@@ -232,7 +232,15 @@ var results = coll.find({
     },{
       'name' : 'Thor'
     }]
-});</code></pre><h3>Features which support 'find' queries</h3><p>These operators can be used to compose find filter objects which can be used within : </p>
+});</code></pre>
+<p><strong>$exists</strong> - filter for documents which contain (even when the value is null) this field or not<p>
+<pre class="prettyprint source lang-javascript"><code>// fetch documents which do not have this field. Use '$exists': true for documents which have this field (it may be null)
+var results = coll.find({
+  'age': {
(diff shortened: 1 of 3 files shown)


plepe:
To be more exact, it also fixes other operators like `$ne` for nested objects:

```js
db.insert({ "foo": "bar", "nested": { "x": true } })
db.find({ "xxx": { $ne: "nothing" }})
-> [ {...} ]

db.find({ "nested.y": { $ne: "nothing" }})
-> []
```

After this change, it will still not return the object, when `nested` is not missing on the main object - maybe we should change this too?

obeliskos:
Looks good, can you provide a unit test in spec/generic/ops.spec.js ?

plepe:
Yes, I will look into it later today or tomorrow.

plepe:
It was faster than expected :-) Should I merge the tests to the commits and force-push?

obeliskos:
Great, thanks!

obeliskos:
Oh I did miss that you modified the html tutorial which is auto generated from the tutorials/Query Examples.md file.  Not a big deal but maybe you can update that in a future checkin.

PR #710, opened 2018-09-27
Precompile $regex filters recursively

plepe:
`find({ "name": { $not: { $regex: '^S' }}})`
would result in a "b.test is not a function" error, because $regex would not
have been compiled.

### DIFF
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -3227,10 +3227,11 @@
           value = new RegExp(value);
         }
       }
-
-      if (typeof value === 'object') {
+      else if (typeof value === 'object') {
         for (var key in value) {
-          value[key] = precompileQuery(key, value[key]);
+          if (key === '$regex' || typeof value[key] === 'object') {
+            value[key] = precompileQuery(key, value[key]);
+          }
         }
       }
 
@@ -3348,7 +3349,9 @@
         throw new Error('Do not know what you want to do.');
       }
 
-      value = precompileQuery(operator, value);
+      if (operator === '$regex' || typeof value === 'object') {
+        value = precompileQuery(operator, value);
+      }
 
       // if user is deep querying the object such as find('name.first': 'odin')
       var usingDotNotation = (property.indexOf('.') !== -1);
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -3328,14 +3328,27 @@
         throw new Error('Do not know what you want to do.');
       }
 
-      // for regex ops, precompile
-      if (operator === '$regex') {
-        if (Array.isArray(value)) {
-          value = new RegExp(value[0], value[1]);
-        } else if (!(value instanceof RegExp)) {
-          value = new RegExp(value);
+      // precompile recursively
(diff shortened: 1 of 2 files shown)


obeliskos:
Thanks, I will need to verify the performance impact of the function call using 'npm run benchmark' later today.  Ideally any added overhead, such as function call would be made only when $regex op exists in the query.

plepe:
I know, that performance is critical, but how to recognize whether a `$regex` operation is present, without iterating through the object? Anyway, the precompile function is called only once per execution, so impact should be minimal.

src/lokijs.js
```diff
@@ -3328,15 +3328,28 @@
...
       }
 
-      // for regex ops, precompile
-      if (operator === '$regex') {
-        if (Array.isArray(value)) {
-          value = new RegExp(value[0], value[1]);
-        } else if (!(value instanceof RegExp)) {
-          value = new RegExp(value);
+      // precompile recursively
+      function precompile (operator, value) {
```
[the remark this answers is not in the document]
plepe: Yes, you are right. Can't drop the function definition though, as I need to call this function recursively ...

plepe:
I moved the precompile function out of the find() function (and renamed it to precompileQuery). Shall I merge these commits into one push force?

src/lokijs.js
```diff
@@ -3217,6 +3217,26 @@
...
+        if (Array.isArray(value)) {
+          value = new RegExp(value[0], value[1]);
+        } else if (!(value instanceof RegExp)) {
+          value = new RegExp(value);
+        }
+      }
+
+      if (typeof value === 'object') {
+        for (var key in value) {
+          value[key] = precompileQuery(key, value[key]);
```
obeliskos: perhaps we could constrain this recursive call to only execute if :
```javascript
if (key === '$regex' || typeof value[key] === 'object')

File: src/lokijs.js [hunk 3]
```diff
@@ -3328,14 +3348,7 @@
...
 
-      // for regex ops, precompile
-      if (operator === '$regex') {
-        if (Array.isArray(value)) {
-          value = new RegExp(value[0], value[1]);
-        } else if (!(value instanceof RegExp)) {
-          value = new RegExp(value);
-        }
-      }
+      value = precompileQuery(operator, value);
```
obeliskos: again, if we could constrain this to only execute when :
```javascript
if (operator === '$regex' || typeof value === 'object') {
```

obeliskos: feel free to discuss any of the above comments, they do however improve results of our single object lookups which have primitive values as well as $eq (non $regex) ops.

src/lokijs.js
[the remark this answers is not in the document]
obeliskos: good suggestion, thanks @Kanaye

plepe:
You are right, I included these suggestions.

obeliskos:
Thanks again @plepe!

plepe:
Great, thanks!

PR #718, opened 2018-10-23
Allow indexing on nested fields

Moocar:
Loki doesn't support indexing on nested fields (period-delimited). This PR adds support for it.

The main change is the addition of the `Utils.getIn` function which performs the deep lookup functionality. Initially, I replaced all `object[path]` lookups with this function, but that resulted in a 2x performance penalty due to the `string.split('.')` that was required for each lookup. So instead, I added a flag `usingDotNotation` that can be set so that the function knows whether it needs to split the path at all.

I've been testing it on https://www.gatsbyjs.org/ as a replacement for the in-memory DB and it's working great so far.

### DIFF
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -97,15 +97,15 @@
           return undefined;
         }
         if (!usingDotNotation) {
-          return object[path]
+          return object[path];
         }
 
         if (typeof(path) === "string") {
-          path = path.split(".")
+          path = path.split(".");
         }
 
         if (!Array.isArray(path)) {
-          throw new Error("path must be a string or array. Found " + typeof(path))
+          throw new Error("path must be a string or array. Found " + typeof(path));
         }
 
         var index = 0,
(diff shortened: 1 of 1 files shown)


obeliskos:
Thanks @Moocar , I have been offline for a while but I hope to look this over soon.

KyleAMathews:
Thanks @obeliskos for all your great work on Loki! Excited to have this in Gatsby soon.

obeliskos:
Great, thanks @Moocar AND @KyleAMathews!  Since this is performance rather than functional related, I will leave option of creating a unit test now or in the future up to you.

Moocar:
@obeliskos thanks! Funny you should mention tests, I just found a bug where nested indexed fields aren't handling the `$in` operator properly. Expect another PR soon :)

retorquere:
Is the use of this documented (preferably a walk-through)?

PR #723, opened 2018-11-29
fix crashing when JSON is incorrect

fudali:
If whole process crashes while storing loki js to the file, the file might be in an error state and it can't be read in the next run. 

The error can't be caught in the application because the error is thrown from JSON.parse inside the callback

### DIFF
--- a/src/loki-fs-structured-adapter.js
+++ b/src/loki-fs-structured-adapter.js
@@ -158,7 +158,11 @@
 
       rl.on('line', function (line) {
         if (line !== "") {
-          obj = JSON.parse(line);
+          try {
+            obj = JSON.parse(line);
+          } catch(e) {
+            callback(e);
+          }
           self.dbref.collections[collectionIndex].data.push(obj);
         }
       });


obeliskos:
sorry, I totally missed this p.r. and am just now looking at it...

I will try to run my stress.js benchmark with and without your change to ensure it doesn't adversely impact performance.... if not i will commit.

So "the whole process crashes" meaning the node vm process crashes?  I assume the crashing point is unrelated to lokijs?  If loki/memory related how big is your db?

fudali:
Hi! Thanks for getting back to it. Yes, by "the whole process crashes" I mean that the node vm process crashes. Maybe I'm missing something but it looks like I can't catch the exception thrown from inside the callback... meaning I can't make sure that by process won't crash without editing lokijs code

obeliskos:
I see no significant performance impact in time or ram usage according to my stress/destress scripts in benchmarks folder.... so I will close and commit.

Thanks!

fudali:
any chance you will release a new version with this fix?

obeliskos:
sorry I realize you made this request before our most recent release, but I would think we should probably wait until we have more fixes before asking @techfort to do another release...

Can you work around this in the near term while we prepare any other fixes which may come up in the next month or so?  @techfort handles all the npm publishing and his free time seems to be little lately.

fudali:
Do you have any advice on how to work around this? Probably I can fork, build and release the lib on my own, right?

PR #738, opened 2019-02-02
remove: avoid "Cannot convert undefined or null to object" error

franck34:
When trying to remove an unexisting doc by his id, this.get() return null. 

But Null is an object so "Parameter is not an object" error is not trigger, 

Then hasOwnProperty.call throw error "Cannot convert undefined or null to object" rather than throw expected error "Object is not a document stored in the collection"

### DIFF
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -6131,7 +6131,7 @@
         return;
       }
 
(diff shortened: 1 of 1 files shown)


franck34:
@techfort hmm do you want a PR with a unit test perhaps ?

PR #743, opened 2019-02-23
Fix firstOnly in branched resultset

### LINKED ISSUE #742: firstOnly not working in branched resultset
`firstOnly` is broken in branched resultsets, i.e. when `this.filterInitialized` is `true`.

Check with the following in the LokiJS playground:
```js
var db = new loki('sandbox.db');
var items = db.addCollection('items');
items.insert({ name : 'mjolnir', owner: 'thor', maker: 'dwarves' });
items.insert({ name : 'gungnir', owner: 'odin', maker: 'elves' });
items.insert({ name : 'tyrfing', owner: 'Svafrlami', maker: 'dwarves' });
items.insert({ name : 'draupnir', owner: 'odin', maker: 'elves' });

var elves = items.addDynamicView('elves')
elves.applyFind({ maker: 'elves' })

var resultset = elves.branchResultset()
// gives 2 results, ignoring firstOnly
var result = resultset.find({ name: { $ne: 'thor' } }, true).data()
logText('Should be one result, but returns two:');
logObject(result)

var resultset2 = elves.branchResultset()
// gives the first item in the collection,
// not the first in the resultset
var result2 = resultset2.find({}, true).data()
logText('Should be the first elf, but is a dwarf:');
logObject(result2)
```

stefanprobst:
Fixes #742

### DIFF
--- a/spec/generic/dynamicview.spec.js
+++ b/spec/generic/dynamicview.spec.js
@@ -291,9 +291,9 @@ describe('dynamicviews', function () {
     });
   });
 
-  describe('querying branched result set', () => {
+  describe('querying branched result set', function () {
     var elves;
-    beforeAll(() => {
+    beforeAll(function () {
       var db = new loki('firstonly.db');
       var items = db.addCollection('items');
       items.insert({ name: 'mjolnir', owner: 'thor', maker: 'dwarves' });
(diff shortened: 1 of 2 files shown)


obeliskos:
Good catch, I am able to see the problem with your example gist in #742 and that does seem to be a bug you have fixed.

Would you be able to add a simple unit test to ensure this doesn't break in the future?

Thanks

stefanprobst:
I added the example gist as tests. not sure if `dynamicview.spec` is the correct file -- if not please advise where best to move them.

obeliskos:
Great! Thank you.

Issue #455, opened 2016-07-21
Object.observe has been deprecated to Proxy API


ohenepee:
https://developers.google.com/web/updates/2016/02/chrome-49-deprecations?hl=en#objectobserve-is-deprecated

I keep getting console warnings from Chrome (haven't tested Firefox) each time I use LokiJS... just wanted to give a heads up on the deprecation before it breaks dependent apps or perhaps doesn't break them for too long.

https://www.chromestatus.com/features/6147094632988672 says `Object.observe` has been removed in Chrome 50 [ personally I use Chrome 49 for development ]

obeliskos:
Sorry late reply, so I guess they don't want to add any features to JS spec that can't be completely polyfilled.  

Notice to @Cortys (re: #231) : I guess we will have to watch this and remove if this becomes breaking.

Cortys:
Sure. It should not become breaking if autoupdates are not enabled though. Chrome only complains because of [this check](https://github.com/runnr/LokiJS/blob/db8b5c49794b2a794638df886163fd4a7f000068/src/lokijs.js#L3007) which is interpreted as "usage" of a deprecated API.

I've thought about reimplementing autoupdates via Proxies, which would also make them synchronous, but that may not be feasible for performance reasons and it also does not play well with the object inflation feature of `loadDatabase`.

Since the `Object.observe` based implementation does not / will not work anymore, removing autoupdates is probably the best thing to do for now. Let me know if you need a pull request reverting #233.

xland:
Object.observe，This feature is obsolete. Try to avoid using it. Use Proxy.

Electron can not auto update collections.

PR #780, opened 2019-05-28
Proxify autoupdate

steven-peralta:
since `Object.observe` has been deprecated, I went ahead and updated the feature to use the Proxy class instead. when `autoupdate` is enabled all documents that are inserted via `Collection.insertOne` or copied via `Loki.loadJSONObject` are proxified with `Collection.autoUpdateProxification`, which will return a document encapsulated by a Proxy or, if `autoupdate` is disabled, the object itself

Fixes issue #455

### DIFF
--- a/src/lokijs.js
+++ b/src/lokijs.js
@@ -4904,32 +4904,6 @@
         this.ensureIndex(indices[idx]);
       }
 
-      /*function observerCallback(changes) {
-
-        var changedObjects = typeof Set === 'function' ? new Set() : [];
-
-        if (!changedObjects.add)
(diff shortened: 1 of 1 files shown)


steven-peralta:
hmm, passes all tests except for the Node 4 test which times out.

is there a reason to keep supporting Node v4 for this project seeing as Node v4 is EOL for a little over a year now?

percy507:
Hi @techfort ,

The `Object.observe` method has been deprecated, do you mind check this PR and switch `Object.observe` into `Proxy`?

Issue #806, opened 2019-09-24
lokijs-indexed-adapter breaks badly on Safari 13


radex:
It seems that Safari 13 has a bug (which I sadly didn't catch while it was in beta) in the way it measures how much space is used by the IndexedDB database. It thinks that much more is in use than there actually is. In [Nozbe 4](http://4.nozbe.com), just by using the app, we could get Safari to ask for 10GB of storage space, even though in reality, <100MB was used (still a lot, but two orders of magnitude less).

I think this has to do with measuring how much space the sqlite write-ahead-log takes up, and so every time Loki database gets saved (so essentially after every database operation), 100MB is added to the measure.

(I'm pretty sure this issue affects all LokiJS databases that are large enough - it surely affects all WatermelonDB databases: https://github.com/Nozbe/watermelondb/ - which use LokiJS under the hood, and I can't think of a reason why it wouldn't affect Loki)

I reported the issue: https://bugs.webkit.org/show_bug.cgi?id=202137 , but given the apparent rarity of the issue: https://twitter.com/bradeeoh/status/1176178498408370176 and knowing Apple's schedules, it will take a while until a fix is shipped.

**And so I'm thinking of alternative solutions to this problem - I'm of course willing to do the work of putting it together, but I could use some advice**

What I'm thinking about -- is it _really_ necessary to save the entire database on every change? that gets very slow on our scale. Partitioning adapter helps a bit, because it saves only the dirty collection, but in our case, just one collection takes up the majority of space.

Is it possible (without reworking the entire library) to save only the dirty records? A loki collection could keep track of which records were added, updated, or deleted, and only those changes would be committed to IDB — plus probably one big record for all the collection metadata

radex:
@techfort @obeliskos Hey, what do you think about this issue? TL;DR is the last paragraph

radex:
#808 is the fix + makes Loki on IndexedDB faster

radex:
closing - use IncrementalIndexedDBAdapter to solve this issue

PR #808, opened 2019-09-26
Incremental IndexedDB adapter

radex:
I'm working towards solving this problem: https://github.com/techfort/LokiJS/issues/806

(TL;DR: To be able to use a large LokiJS database that frequently saves on Safari 13, saving the entire database in one big blob just won't do… and current partitioning implementation doesn't help either because in my case most of that large database is in just one big collection)

### The plan

1. Add lightweight changes tracking - MUCH more lightweight than current Changes API, only track which records ($loki IDs) were added/updated, and which deleted
2. Add a new `IncrementalIndexedDBAdapter` (or better yet, a generic `IncrementalAdapter`, similar to `PartitioningAdapter`):
  - Each record (ideally) is stored separately as separate values in IDB
  - Database metadata (or per-collection metadata) is stored separately
  - On save, adapter checks which records changed, and only saves those + metadata

### Measurements

DB size:

- For my app, the entire database is ~45MB when encoded as JSON (90MB encoded as UTF-16, which is what Safari does to IDB)
- So it takes only 11 saves to reach 1GB of disk space wasted by Safari WAL file (triggering #806)
- Metadata (everything, including indexes, other than collection.data) is only 1.1MB (i think still 1.1MB in Safari VM, because it's entirely ASCII and I think JSC opts for a more compact representation then) -- so probably not even necessary to split metadata per-collection - would take between 440-900 saves to trigger the Safari bug

IDB performance:

- Sadly, it doesn't seem possible to save each record separately - saving new records in IDB is ridiculously slow… saving 65K values (~200B each) took ~3000-4000ms on my computer
  - (for comparison: ~350ms to save one big stringified JSON blob in a single record)
- chunking seems necessary (sad, because it makes adapter more complicated):  650 chunks, 100 elements each takes 250-300ms — faster/about as fast than one big JSON value
- modifying just one chunk takes 12-20ms
  - I suspect most saves in my use case will only touch a single chunk, so that will be a huge performance win, not even taking into account fixing Safari storage bug
- worst case scenario: replacing all chunks - 210-240ms
- loading&parsing all 65K chunks: 200ms
- Chrome: roughly similar numbers, slightly higher than Safari
- Firefox: biggest win - saving one big blob takes 350-400ms (higher than Safari), but saving in 650 chunks only takes ~125ms

## TODO:

- [x] proof of concept implementation
- [x] write tests…
- [ ] polish the implementation


radex:
Update: a working proof of concept implementation is available in `demos/indexed_tests.html` (the code is very dirty at the moment, but works, and is fast!)

for 65K record in a single collection, I get ~300ms for full (initial) save, ~200ms for full load, and ~35ms for incremental save with 80 scattered changes (in real world scenario it would probably be less because changes would be less scattered)

radex:
**Update:**

- The implementation seems to work pretty great!
- I added tests and a browser benchmark runner
- Partly cleaned up code (error handling, etc.)
- The code is still full of console logs, etc… will remove once I'm confident the implementation is solid and doesn't have major bugs
- Sadly, PhantomJS doesn't support IndexedDB or Set, so I don't know how to set it up for automatic tests on CI… I added a quick&dirty tests runner in spec/incrementalidb.html… which is lame, I know, but I don't want to spend too much time figuring out how to set up a more modern browser test runner like Cypress or something

radex:
Quick & dirty benchmark comparison between Safari, Chrome, and Firefox

techfort:
Awesome @radex, thanks for this

PR #820, opened 2019-11-26
Make id field configurable per collection

jbemmel:
The default '$loki' attribute for object ids may work in some cases, but at times it is necessary to specify a different field name to use, for example to achieve compatibility with existing data models or other backends like MongoDB.
This PR makes the name of the ID field configurable per collection, still defaulting to '$loki' but enabling the use of custom field names like '_id'.
npm test passes, one additional test case added.

### DIFF
--- a/spec/generic/collection.spec.js
+++ b/spec/generic/collection.spec.js
@@ -306,7 +306,7 @@ describe('collection', function () {
 
   it('allows to specify a custom ID field name and return instances correctly', function() {
     var db = new loki('test_id.db');
-    var coll = db.addCollection('testcoll', { clone:true, idField: '_id' } );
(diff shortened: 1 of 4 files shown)


spec/generic/collection.spec.js
```diff
@@ -304,4 +304,44 @@ describe('collection', function () {
     expect(obj.b).toEqual(3);
   });
 
+  it('allows to specify a custom ID field name and return instances correctly', function() {
+    var db = new loki('test_id.db');
+    var coll = db.addCollection('testcoll', { clone:true, idField: '_id' } );
```
Sasuke214:
```suggestion
    var coll = db.addCollection('testcoll', { clone: true, idField: '_id' } );
```

jahangirabd:
Useful parameter. Please consider to merge.

PR #827, opened 2020-02-07
capture async errors

DaveWelling:
Currently, any error inside the success handler for the indexeddb key lookup results in the database not loading, but the loadDatabase method (or autoload callback) returning successfully (so that the consuming method thinks everything is ok).
With these changes, I just tried to pass the error up through the callbacks, so I will know something is wrong when loadDatabase fails.


DaveWelling:
Are contributions unwelcome or ...?

Issue #839, opened 2020-05-28
Unique Indices will fail for some user input


radex:
because of this:

```js
    function UniqueIndex(uniqueField) {
      this.keyMap = {};
    };
    /* snip */
    UniqueIndex.prototype.get = function (key) {
      return this.keyMap[key];
    };
```

`collection.by(key, xxx)` cannot have any unsanitized `xxx` input or it might fail. For example, `this.keyMap['hasOwnProperty']` will always be truthy.

One simple solution is to use `Object.create(null)` instead of `{}`, so that the object has no prototype.

radex:
no, bot

PR #847, opened 2020-06-29
Column comparisons & Fix unique index bug

### LINKED ISSUE #636: Column comparisons
I want to make a query that matches all records where, say, `record.oneColumn > record.otherColumn`. This can be useful, for instance, when having two columns with timestamps and you want to fetch records where "last change is newer than last seen".

Currently, I believe it's only possible using `where` queries, but that's undesirable for me because:

- it's tricky when Loki is moved to separate web worker (can't pass functions between threads)
- harder to test (when you have, say, your own model methods/helper functions that you want to check if they produce the right Loki queries)
- it's hard to mix with AND/OR queries

I'm thinking of a declarative syntax, say:

```
.find({ column1: { $gt: { $column: 'column2' } } })
```

or:

```
.find({ column1: { $column: [
  'column2',
  { $eq: 'column2' }
] } })
```

or something of that sort.

**I'm willing to dig into the code and implement the feature myself.** … but I could use a few pointers as to how to approach the problem, and where to start.

radex:
Fixes #636.

This introduces a new class of filters, that do comparisons on document level, instead of on property level.

The key use case is allowing column comparisons, e.g. fetch all tasks where task.foo > task.bar. The syntax is this:

```js
// simple column comparison
{ foo: { $$gt: 'bar' } } 

// for more flexibility, function can be passed
{ foo: { $$gt: doc => doc.bar + 1 } }
```

I took care not to affect hot path's performance. So I don't do any `$$` checks — instead, all LokiOps now take a third argument (document) that's passed around - but only $$ops actually use it.

Nested filters - $or, $and, $not, $len, $size, $elemMatch all work.

I'm also thinking of adding object-level $filter `.find({ $filter: (doc) => xxx })` and extending property-level $filter `.find({ foo: { $filter: (foo, doc) => xxx }  })` for object-level comparisons without having to completely resort to using `.where()`. But I didn't do it in this PR.

Additionally, I fix https://github.com/techfort/LokiJS/issues/839 here. Plain objects {} are not safe for use with arbitrary keys. Since Loki doesn't do ES6, I used prototype-less Objects to fix the issue.

### DIFF
--- a/spec/generic/ops.spec.js
+++ b/spec/generic/ops.spec.js
@@ -428,4 +428,23 @@ describe("Individual operator tests", function() {
       }
     }).length).toBe(1)
   });
+
+  it('$$op column comparisons work', function () {
+    var db = new loki('db');
+    var coll = db.addCollection('coll');
+
+    coll.insert({ a: null, b: 5 });
+    coll.insert({ a: '5', b: 5 });
+    coll.insert({ a: 5, b: 5 });
+    coll.insert({ a: 6, b: 5 });
+    coll.insert({ a: 3, b: 5 });
+    coll.insert({ a: 3, b: 'number' });
+
+    expect(coll.find({ a: { $$eq: 'b' } }).length).toEqual(1);
+    expect(coll.find({ a: { $$aeq: 'b' } }).length).toEqual(2);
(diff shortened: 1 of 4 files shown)


radex:
I hope so! I'm shipping this in production for the last couple of weeks and nothing broke :)

PR #892, opened 2021-08-19
feat: Check if an operator is valid

Losses:
if not valid throw an error to let users know

Current error is: `TypeError: fun is not a function`

This is not straightforward enough


Losses:
bump

Losses:
bump

PR #896, opened 2021-09-29
added missing null check to dot sub scan fixes #891

### LINKED ISSUE #891: Querying objects with dot notation fails with error: TypeError: Cannot use 'in' operator to search for 'b' in null
Querying objects with dot notation fails if there is an object with `null` value 

**Steps to reproduce:**
```javascript
const Loki = require('lokijs');
const db = new Loki();

const coll = db.addCollection('myCollection');

coll.insert({ a: { b: 'c' } });
coll.insert({ a: null });

coll.find({ 'a.b': 'c' });
```

**Error:**
```
/node_modules/lokijs/src/lokijs.js:418
      if (typeof root === 'object' && path in root) {
                                           ^

TypeError: Cannot use 'in' operator to search for 'b' in null
```

**LokiJS version - 1.5.12**

blaky:
When a filter was deep dot filter is applied on null value, the code throws an exception. This is due to a missing null check. `typeof null` is always `object` which is a questionable part of JavaScript and leads to bugs like this...

### DIFF
--- a/spec/generic/ops.spec.js
+++ b/spec/generic/ops.spec.js
@@ -359,6 +359,17 @@ describe("Individual operator tests", function() {
     expect(coll.find({ "c": { $eq: undefined } }).length).toEqual(4);
   });
 
+  it('query nested documents with nullable object', function() {
+    var db = new loki('db');
+    var coll = db.addCollection('coll');
+
+    coll.insert({ a: null, b: 5, c: { a: 1 }});
+    coll.insert({ a: "11", b: 5, c: { a: 1 }});
+    coll.insert({ a: "11", b: 5, c: null});
+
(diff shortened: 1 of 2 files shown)


blaky:
Hi @techfort , 
Could you please check this PR and let me know if there is anything else we can do to get it merged.
Thanks!

techfort:
Damn gotta review this, sorry for the delay

bastiion:
@techfort it is a very simple fix (in only one line of code + test case), could you please merge this PR ? tx

blaky:
@techfort Any chance that you review this in the foreseeable future?

blaky:
@techfort Thank you! :-)

PR #918, opened 2022-11-22
Fix disableFreeze flag

vasilistert2:
The flag `disableFreeze` is in `collection` not in `this`. Now it returns `undefined` so the object is always cloned.


AntonNevzgodov:
Until this PR is merged (never, I guess), this monkeypatch can be a viable solution:
```ts
// TypeScript
import LokiJS from 'lokijs';

// Make Resultset.disableFreeze a property with getter that returns Collection.disableFreeze property value.
Object.defineProperty((LokiJS as any).Resultset.prototype, 'disableFreeze', {
  get() {
    return this.collection.disableFreeze;
  },
});
```

wilwade:
Just hit this bug. Can confirm the PR fixes it.

